PR #15, opened 2019-05-29
Motif redesign

msenejoa:
Redesigned Motif landing page to include a sticky header with a motif toggle, fixed issue with storybook showing `object Object` when expecting raw html


pzhine:
Please bump the version to 1.0.6

PR #18, opened 2019-05-31
Greybox border

msenejoa:
To be added after PR15 with v1.0.6. Added a bottom border to the motif header that appears once the content header is not visible on the scroll box. 

Added a story to the Motif Page with a long title and longer body


msenejoa:
To be added after PR21 with v1.0.6. Added a bottom border to the motif header that appears once the content header is not visible on the scroll box.

Added a story to the Motif Page with a long title and longer body

PR #25, opened 2019-06-06
Motif Redesign

msenejoa:
With Sticky Header


packages/databyss-ui/modules/Landing/Landing.js
```diff
@@ -1,48 +1,128 @@
...
+    super(props)
+    this.state = {
+      borderBottom: false,
+      contentLoaded: false,
+      headerLoaded: false,
+    }
+  }
+
+  componentDidMount() {
+    window.addEventListener('scroll', this.handleScroll.bind(this), true)
```
pzhine: please throttle this so that it doesn't call on every scroll event

packages/databyss-ui/modules/Landing/Landing.js
```diff
@@ -1,48 +1,128 @@
...
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('scroll', this.handleScroll.bind(this), true)
+  }
+
+  handleScroll() {
+    // check to see both header and content header are mounted
+    const { contentLoaded, headerLoaded, borderBottom } = this.state
+    if (document.getElementById('landing-header') && !headerLoaded) {
```
pzhine: rather than using `getElementById`, use `this.headerRef.current` (see comment below)

packages/databyss-ui/modules/Landing/Landing.js
```diff
@@ -1,48 +1,128 @@
...
+      subtitle,
+      showMotifLinks,
+      onMotifLinksChange,
+      withToggle,
+    } = this.props
+
+    return (
+      <Content className={classnames(className, classes.landing)}>
+        <LandingHeading
+          id="landing-header"
```
pzhine: rather than use a DOM id (which is static and will fail if more than one `<Landing />` is on a page), use `ref={this.headerRef}` (see comment in constructor)

packages/databyss-ui/modules/Landing/Landing.js
```diff
@@ -1,48 +1,128 @@
...
-            React.cloneElement(renderCfItem(cf), { key: cf.id })
+class Landing extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      borderBottom: false,
+      contentLoaded: false,
+      headerLoaded: false,
+    }
+  }
```
pzhine: here is where you instantiate the header ref, like `this.headerRef = React.createRef()`

packages/databyss-ui/components/Heading/HeadingContainer.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import classnames from 'classnames'
+import injectSheet from 'react-jss'
+import styles from './styles'
+
+const LandingHeading = ({ classes, className, children, style }) => (
```
pzhine: There's a naming mismatch here. Did you mean `const HeadingContainer =`?

packages/databyss-ui/stories/Landing/motifLanding.stories.js
```diff
@@ -22,12 +22,23 @@ const landingProps = {
     </Link>
   ),
 }
+const landingPropsLong = {
+  cfList,
+  title:
+    'acques Derrida on HIEROGLYPH / PASIGRAPHY / LEIBNIZ / CHINESE / UNIVERSAL LANGUAGE',
```
pzhine: Fix typo plz

packages/databyss-ui/modules/Landing/Landing.js
msenejoa: for future reference `react-jss` breaks `React.forwardRef` because it removes the `ref` component. Found a solution here [here](https://github.com/cssinjs/react-jss/issues/277)

PR #28, opened 2019-07-03
Senejoa/sources page

msenejoa:
To correspond with PR #41 on returntocinder

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "databyss-org-workspace",
-  "version": "1.1.0",
+  "version": "1.0.10",
   "description": "massive open-source citation database platform",
   "main": "index.js",
   "repository": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "databyss-org-workspace",
-  "version": "1.0.9",
+  "version": "1.1.0",
   "description": "massive open-source citation database platform",
   "main": "index.js",
   "repository": {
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "databyss-org-workspace",
-  "version": "1.0.9",
+  "version": "1.1.0",
```
pzhine: The version should be `1.0.10`. We'll use `1.1.x` as the next major release (with the primitives, etc).

packages/databyss-ui/components/Heading/PageSubHeading.js
```diff
@@ -10,20 +10,30 @@ const PageSubHeading = ({
...
-        : classes.headerPageSubHeadingSticky
-    )}
-    style={style}
-  >
-    {children}
-  </div>
-)
+  sourceHeader,
+}) => {
+  const motifOrSourceSticky = !sourceHeader
```
pzhine: Do these classes need to be different? It's better if we can unify them.

packages/databyss-ui/modules/Landing/Landing.js
```diff
@@ -145,10 +145,13 @@ class Landing extends React.Component {
       showMotifLinks,
       onMotifLinksChange,
       withToggle,
+      isSource,
     } = this.props
 
     const { headerSticky, inStickyContainer, isViewMobile, width } = this.state
 
+    const sourceHeader = isSource
```
pzhine: Again, let's try to avoid this. If you need to fork the styles or behavior, refactor to `Landing`, `SourceLanding` and `MotifLanding` using a composition pattern (cf `Button`, `PrimaryButton`, etc).

pzhine:
> Need more whitespace here:

Screenshot 2019-07-03 22 19 01

> We should do this by adjusting the margin/padding of these components so they look the same on Motif Landing:

Screenshot 2019-07-03 22 20 25

PR #56, opened 2019-09-06
michael/editor-api-integration

pzhine:


packages/databyss-services/pages/index.js
```diff
@@ -0,0 +1,28 @@
+import { httpGet, httpPost } from '../lib/requestApi'
+
+// TODO: Add native versions of these
+export function setAccountId(value) {
```
pzhine: The accountId get/set functions should live in `databyss-services/auth/index.js` alongside the the token get/set functions

packages/databyss-ui/components/Editor/SlateContentEditable.js
```diff
@@ -22,16 +26,32 @@ const toSlateJson = (editorState, pageBlocks) => ({
...
   onActiveBlockContentChange,
   onEditableStateChange,
+  onNewBlock,
+  onBackspace,
 }) => {
   const [editorState] = useEditorContext()
+
+  // const { activeBlockId, editableState, blocks, page } = editorState
   const { activeBlockId, editableState, blocks, page } = editorState
   const editableRef = useRef(null)
```
pzhine: Doesn't seem like we're using this rn, so let's remove it

packages/databyss-ui/components/Editor/state/reducer.js
```diff
@@ -40,14 +42,15 @@ export const setRawHtmlForBlock = (state, block, html) => {
...
-const setActiveBlockType = (state, type, isNew) => {
+const setActiveBlockType = (state, type, isNew, _id) => {
   // changing block type will always generate a new refId
   const nextRefId = ObjectId().toHexString()
   const activeBlock = state.blocks[state.activeBlockId]
   const rawHtml = isNew ? '' : getRawHtmlForBlock(state, activeBlock)
   const nextState = cloneDeep(state)
   nextState.blocks[state.activeBlockId] = {
     ...nextState.blocks[state.activeBlockId],
+    _id,
```
pzhine: Why do you need to specify the `_id` here? Can't we use the existing one?

packages/databyss-ui/components/Editor/state/reducer.js
```diff
@@ -64,6 +67,74 @@ const setActiveBlockType = (state, type, isNew) => {
   }
 }
 
+const insertBlock = (state, payload) => {
+  const {
+    activeBlockId,
```
pzhine: Let's rename these 2 `insertedBlockId` and `insertedBlockHtml` so they're not confused with `state.activeBlockId`

Then we should add an invariant assertion like:
`invariant(insertedBlockId === state.activeBlockId, 'insertedBlockId must match activeBlockId. It's possible that you called insertNewActiveBlock before activeBlockId was updated')`

packages/databyss-ui/components/Editor/state/reducer.js
```diff
@@ -64,6 +67,74 @@ const setActiveBlockType = (state, type, isNew) => {
   }
 }
 
+const insertBlock = (state, payload) => {
```
pzhine: Let's rename this `insertNewActiveBlock` for clarity and make the args clearer:
`const insertNewActiveBlock = (state, { insertedBlockId, insertedBlockHtml, previousBlockId, previousBlockHtml }) { ... }`

packages/databyss-ui/components/Editor/state/reducer.js
```diff
@@ -64,6 +67,74 @@ const setActiveBlockType = (state, type, isNew) => {
...
+  const {
+    activeBlockId,
+    activeBlockText,
+    previousBlockId,
+    previousBlockText,
+  } = payload.blockProperties
+
+  let _state = state
```
pzhine: This smells bad. It implies that `_state` and `state` point to different objects, when they don't. You should change it to `_state = cloneDeep(state)` because you are changing the `.page.blocks` leaf "in-place" on 79

packages/databyss-ui/components/Editor/state/reducer.js
pzhine: I'll do this one

packages/databyss-ui/components/Editor/state/reducer.js
pzhine: ignore

packages/databyss-ui/components/Editor/state/reducer.js
pzhine: i'll do this one

packages/databyss-ui/components/Editor/state/reducer.js
pzhine: i'll do this one

PR #67, opened 2019-09-12
topic tag

msenejoa:


packages/databyss-api/src/routes/api/helpers/pagesHelper.js
```diff
@@ -9,14 +10,17 @@ const getBlockItemsFromId = blocks => {
       _id,
     }).catch(err => console.log(err))
     if (block) {
-      const { type, entryId, sourceId, authorId } = block
+      const { type, entryId, sourceId, authorId, topicId } = block
       const response = { type, _id }
       if (type === 'ENTRY') {
```
pzhine: a more functional way to do this would be:
```
response.refId = {
  ENTRY: entryId,
  SOURCE: sourceId,
  TOPIC: topicId,
  AUTHOR: authorId,
}[type]
```

packages/databyss-api/src/routes/api/helpers/pagesHelper.js
```diff
@@ -33,19 +37,37 @@ const getSourcesFromId = list => {
...
+    if (!source) return {}
+
     const { resource } = source
     const response = { rawHtml: resource, _id }
     return response
   })
   return Promise.all(promises)
 }
 
+const getTopicsFromId = list => {
```
pzhine: This is almost a duplicate of getEntriesFromId. Let's refactor to a more generic form, like:
```
const populateRefEntities = (list, Model) =>
  Promise.all(list.map(async b => {
    const _id = b.refId
    const entity = await Model.findOne({ _id })
    if (!entity) { throw new BadRefIdError(refId) }
    return { rawHtml: entity.text }
  }))
```
Note that we should create a special error type `BadRefIdError` so that we can monitor these in the logs and fix data integrity problems.

Also note that this requires normalizing the entity models (Entry, Topic, Source, etc) to have a `text` key that has the text of the entity.

packages/databyss-api/src/routes/api/pages.js
```diff
@@ -105,6 +142,9 @@ router.post(
           if (type === 'ENTRY') {
             blockFields.entryId = refId
           }
+          if (type === 'TOPIC') {
```
pzhine: This can be more concise than a sequence of `if` blocks, see comment above.

packages/databyss-api/src/routes/api/pages.js
```diff
@@ -209,11 +249,16 @@ router.get(
       let entries = await getEntriesFromId(entriesList)
       entries = dictionaryFromList(entries)
 
+      const topicsList = blockList.filter(b => b.type === 'TOPIC')
```
pzhine: If we refactor to the more generic `populateRefEntities`, we can rewrite this more concisely:
```
const [sources, entries, topics] = ['SOURCE', 'ENTRY', 'TOPIC'].map(t => {
  const list = blockList.filter(b => b.type === t)
  const populated = populateRefEntities(list, t)
  return dictionaryFromList(populated)
})
```

packages/databyss-ui/components/Editor/slate/ContentEditable.js
```diff
@@ -14,7 +14,7 @@ const toSlateJson = (editorState, pageBlocks) => ({
   document: {
     nodes: pageBlocks.map(block => {
       const textBlock =
-        block.type === 'SOURCE'
+        block.type === 'SOURCE' || block.type === 'TOPIC'
```
pzhine: It seems like we're repeating this OR a lot. Let's replace all with a function like `isAtomicInlineType(type)`

packages/databyss-ui/components/Editor/state/reducer.js
```diff
@@ -33,11 +34,15 @@ const cleanUpState = state => {
     if (type === 'ENTRY') {
       _entries[refId] = entries[refId]
     }
+    if (type === 'TOPIC') {
```
pzhine: We're also repeating ourselves a lot when we are accessing entities in the state. How about a selector like
```
const entities = (state, type) => {
  ENTRY: state.entries,
  SOURCE: state.sources,
  TOPIC: state.topics,
  AUTHOR: state.authors,
}[type]
```

Then we can write this as:
```
pageBlocks.forEach(b => {
  const { type, refId } = blocks[b._id]
  entities(state, type)[refId] = entities(state, type)[refId]
}
```

packages/databyss-ui/components/Editor/state/reducer.js
```diff
@@ -47,6 +52,8 @@ export const getRawHtmlForBlock = (state, block) => {
       return state.entries[block.refId].rawHtml
     case 'SOURCE':
       return state.sources[block.refId].rawHtml
+    case 'TOPIC':
```
pzhine: With the selector:
```
const getRawHtmlForBlock = (state, block) => entities(state, block.type).rawHtml
```

packages/databyss-api/src/routes/api/helpers/pagesHelper.js
msenejoa: `const entity = await eval(Model).findOne({ _id })`
gives me the error that Source, Entry, Topic are used but never defined when running tests

PR #70, opened 2019-09-18
Michael/bold italic

msenejoa:


packages/databyss-ui/components/Editor/slate/ContentEditable.js
```diff
@@ -3,38 +3,52 @@ import { KeyUtils, Value } from 'slate'
 import ObjectId from 'bson-objectid'
 import { Editor } from 'slate-react'
 import EditorBlock from '../EditorBlock'
-import EditorInline from '../EditorInline'
+// import EditorInline from '../EditorInline'
```
pzhine: shouldn't be commented out

packages/databyss-ui/components/Editor/slate/ContentEditable.js
```diff
@@ -48,12 +62,17 @@ const schema = {
...
+  const isSelected = editor.value.selection.focus.isInNode(node)
+  const style = isSelected
+    ? {
+        backgroundColor: '#efefef',
+      }
+    : {}
   if (node.type === 'SOURCE') {
     return (
-      <EditorInline isFocused={isFocused} {...attributes}>
+      <span style={style} {...attributes}>
```
pzhine: This should be `EditorInline`

packages/databyss-ui/components/Editor/state/reducer.js
```diff
@@ -154,6 +156,35 @@ const backspace = (state, payload) => {
   return cleanUpState(_state)
 }
 
+const getMarkupValues = (nextState, blockValue) => {
```
pzhine: We shouldn't be doing the conversion of `stateToSlate` or `slateToState` anywhere in the `state/reducer`. By this point the marks should be converted to a `ranges` key on the payload

packages/databyss-ui/components/Editor/slate/markup.js
```diff
@@ -0,0 +1,78 @@
+import cloneDeep from 'clone-deep'
+import { Editor, Value } from 'slate'
+
+export const stateToSlate = (slate, _id) => {
```
pzhine: I think you have these named backwards...this one should be `slateToState`, as in, `slate block` => `state entity`. Maybe a better name would be slateBlockToStateEntity

packages/databyss-ui/components/Editor/slate/markup.js
```diff
@@ -0,0 +1,78 @@
...
+  const { nodes } = slate
+  let text = ''
+  let ranges = nodes.map((n, i) => {
```
pzhine: Let's extract this to a fn called `getRangesFromBlock` so that we can use it here and in `ContentEditable`

packages/databyss-ui/components/Editor/slate/ContentEditable.js
```diff
@@ -111,13 +143,17 @@ const SlateContentEditable = ({
     const _nextText = _nextEditableState.value.document.getNode(activeBlockId)
       .text
     if (_nextText !== _prevText) {
-      onActiveBlockContentChange(_nextText, _nextEditableState)
+      const block = _nextEditableState.value.anchorBlock
+      const jsonBlockValue = { ...block.toJSON(), key: block.key }
+
+      onActiveBlockContentChange(_nextText, _nextEditableState, jsonBlockValue)
```
pzhine: We shouldn't be passing Slate JSONs outside of the ContentEditable component, because it breaks our abstraction. The 3rd argument should be `ranges`, the return value of our `getRangesFromBlock` that we'll create in `components/Editor/slate/markup.js`

packages/databyss-ui/components/Editor/state/reducer.js
```diff
@@ -75,11 +69,50 @@ export const setRawHtmlForBlock = (state, block, html) => {
   return nextState
 }
 
+export const correctRangeOffsetForBlock = (state, block, offset) => {
+  const _state = cloneDeep(state)
+  switch (block.type) {
```
pzhine: Don't need the switch, can use  
```
const refEntity = entities(_state, block.type)[block.refId]
refEntity.ranges = refEntity.ranges.map(r => ({ ...r, offset: r.offset + offset }))
```

BUT

looks like we need to get that refEntity below in `getRangesForBlock`, so let's make a fn `getBlockRefEntity(state, block)` so then we can just do:
```
getBlockRefEntity(_state, block).ranges = getBlockRefEntity(_state, block).ranges.map(r => ({ ...r, offset: r.offset + offset }))
```

packages/databyss-ui/components/Editor/state/reducer.js
```diff
@@ -75,11 +69,50 @@ export const setRawHtmlForBlock = (state, block, html) => {
...
+        r => ({ ...r, offset: r.offset + offset })
+      )
+      return _state
+    default:
+      throw new Error('Invalid block type', block.type)
+  }
+}
+
+export const getRangesForBlock = (state, block) => {
+  switch (block.type) {
```
pzhine: We can just return `getBlockRefEntity(state, block).ranges` here

PR #71, opened 2019-09-19
Render inlines to static HTML

pzhine:


pzhine:
@msenejoa merged these into michael/atomic-markup

PR #73, opened 2019-09-30
Michael/atomic markup

msenejoa:


packages/databyss-ui/components/Editor/slate/hotKeys.js
```diff
@@ -1,14 +1,25 @@
 import { isKeyHotkey } from 'is-hotkey'
 import { IS_IOS, IS_MAC } from 'slate-dev-environment'
+
+export const IS_LINUX =
```
pzhine: Let's move this to `databyss-ui/lib/dom.js` and import it here and in `cypress/support/commands.js` (rather than it being a parameter on every command).

packages/databyss-ui/components/Editor/state/reducer.js
```diff
@@ -94,6 +96,24 @@ export const getRangesForBlock = (state, block) => {
   }
 }
 
+export const setRangesForBlock = (state, block, ranges) => {
+  const nextState = cloneDeep(state)
+  switch (block.type) {
```
pzhine: Can use `entities(state, block.type)[block.refId].ranges = ranges`

PR #96, opened 2019-10-31
Michael/content editable refactor

msenejoa:


packages/databyss-ui/components/Editor/slate/line/ContentEditable.js
```diff
@@ -0,0 +1,110 @@
...
+
+  const onKeyDown = (event, editor, next) => {
+    if (event.key === 'Enter') {
+      editor.insertText('\n')
+      return event.preventDefault()
+    }
+
+    editorHotKeys(event, editor, onHotKey, OnToggleMark)
+
+    if (hotKeys.isLocation(event)) {
```
pzhine: Let's leave locations out of line for now

packages/databyss-ui/components/Editor/slate/line/ContentEditable.js
```diff
@@ -0,0 +1,110 @@
...
+    // if markup is applied to selection. update state range
+    return false
+  }
+
+  const onChange = change => {
+    const { value } = change
+    if (onDocumentChange) {
+      onDocumentChange(value.document.toJSON())
+    }
+    const blockChanges = checkActiveBlockContentChanged({ value })
```
pzhine: Do we really need `checkActiveBlockContentChanged`? I know we don't want to fire `onChange` unless the text has changed, but isn't there a more efficient check for this in the single-line context? 
At the very least, we need to change the name of this fn to avoid confusion.

packages/databyss-ui/components/Editor/EditorBlock.js
```diff
@@ -64,21 +65,23 @@ const EditorBlock = ({ children, node }) => {
...
-        overflow="visible"
-      >
-        {node.text.length < 1 && (
-          <EditorBlockMenu
-            hideCursor={bool => setMenuActive(bool)}
-            node={node}
-          />
-        )}
-      </View>
+      {node.type !== 'TEXT' && (
```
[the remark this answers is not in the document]
msenejoa: Ive created a component called `EditorLine` inside of `EditorBlock` thats made for single line use

packages/databyss-ui/Editor/EditorLine.js
```diff
@@ -37,7 +37,13 @@ const EditorLine = ({ children, onStateChange, value }) => {
   }
 
   // should only have 1 child (e.g. draft/ContentEditable or slate/ContentEditable)
+
+  const _css = children.props.css ? [...children.props.css] : null
+
+  // TODO: CSS GETS OVERRIDDEN HERE
   return React.cloneElement(React.Children.only(children), {
+    ...children.props,
+    _css,
```
pzhine: I encountered this problem with the `css` prop in another context, so it's not isolated to this component tree. Let's use this standard for props:
`overrideCss`: css style object that replaces the `css` prop on the child
`appendCss`: css style object or array of objects that get appended to the css or css array on the child

so an implementation on the child would look like:
```jsx
const _css = { position: 'absolute' }
const Modal = ({
  children,
  overrideCss,
  concatCss,
  ...others
}) => (
  <StyledReactModal
    css={overrideCss || [_css].concat(concatCss)}
    {...others}
  >
    {children}
  </StyledReactModal>
)
```

packages/databyss-ui/Editor/EditorPage.js
```diff
@@ -40,18 +42,27 @@ const EditorPage = ({ children }) => {
   }
 
   const onBlockBlur = (id, text, editableState) => {
-    // check if block is empty
     // if empty replace block with virgin block
     if (text.length === 0 && id) {
-      // check block to see if ranges exist
-      dispatchEditor(clearBlock(id, editableState))
+      if (!isBlockEmpty(id, editableState)) {
```
pzhine: I don't understand this case. How can `text.length` be 0 but block not empty?

packages/databyss-ui/Editor/EditorPage.js
```diff
@@ -40,18 +42,27 @@ const EditorPage = ({ children }) => {
...
+        dispatchEditor(clearBlock(id, editableState))
+      }
     }
 
     if (text.trim().match(/^@/) && editorState.blocks[id].type !== 'SOURCE') {
-      onSetBlockType('SOURCE', id, editableState)
+      // if only @ or # was entered
+      if (text.trim().length === 1) {
```
pzhine: As this condition and action is the same for more than one block type, coalesce into one condition (if match @ or # and length is 1)

packages/databyss-ui/Editor/slate/line/ContentEditable.js
```diff
@@ -0,0 +1,132 @@
...
+const SlateContentEditable = forwardRef(
+  (
+    {
+      onContentChange,
+      onEditableStateChange,
+      onNativeDocumentChange,
+      OnToggleMark,
+      onHotKey,
+      onBlur,
+      _css,
```
pzhine: change to `overrideCss` as discussed above

packages/databyss-ui/Editor/slate/line/ContentEditable.js
```diff
@@ -0,0 +1,132 @@
...
+    // solution from https://itnext.io/reusing-the-ref-from-forwardref-with-react-hooks-4ce9df693dd
+    const combinedRef = useCombinedRefs(ref, editableRef)
+
+    // checks editor state for active block content changed
+    const checkContentChanged = _nextEditableState => {
+      // on first click on change returns null values for anchor block
+      if (!_nextEditableState.value.anchorBlock) {
+        return false
+      }
+      // if (!textValue) {
```
pzhine: cleanup scratch

packages/databyss-ui/Editor/EditorPage.js
msenejoa: this dispatches a helper function which will replace the current block with a new block. blocks can have left over `marks` or have a `type` which isn't `ENTRY`

PR #114, opened 2019-11-14
Michael/sources provider

msenejoa:
* Created Sources provider and wrapper
* Loading Component
* Account//Sources// Load and Save story (must be authorized)
* Error Component
* Sources Express route


packages/databyss-api/src/models/Source.js
```diff
@@ -14,21 +14,24 @@ const SourceSchema = new mongoose.Schema({
...
-    required: true,
+    // required: true,
   },
-  ranges: { type: Array },
-  citations: {
+  name: {
     type: String,
+    //  required: true,
   },
+  ranges: { type: Array },
```
pzhine: Ranges applies to a single citation, i.e. each citation in `citations` has a `ranges` field.
I think we can use mongoose [subdocuments](https://mongoosejs.com/docs/subdocs.html) for this. So we'd have a `richText` schema: `const RichTextSchema = new Schema({ textValue: { type: String }, ranges: { type: Array })` and then in the source schema: `citations: [RichTextSchema]`.

We should do the same with `ranges`, as in declare a `RangeSchema` that defines the fields of a range.

packages/databyss-api/src/routes/api/sources.js
```diff
@@ -0,0 +1,79 @@
...
+    try {
+      const source = await Source.findOne({
+        _id: req.params.id,
+      })
+
+      if (!source) {
+        return res.status(404).json({ msg: 'There is no source for this id' })
+      }
+
+      if (req.account.id.toString() !== source.account.toString()) {
```
pzhine: Isn't this handled by `accountMiddleware`?

packages/databyss-services/sources/ShowFirstCitation.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
```
pzhine: This component should be moved to the story as it's for demo/testing purposes

packages/databyss-services/sources/SourceNotFoundError.js
```diff
@@ -0,0 +1,15 @@
+export class SourceNotFoundError extends Error {
+  constructor(message, sourceId) {
+    super()
+    this.message = message
+    this.sourceId = sourceId
+  }
+}
+
+export class NotAuthorized extends Error {
```
pzhine: This doesn't seem to be used anywhere

packages/databyss-services/sources/SourcesProvider.js
```diff
@@ -0,0 +1,72 @@
...
+    </View>
+  </View>
+)
+
+const SourceProvider = ({ children, initialState, reducer }) => {
+  const [state, dispatch] = useReducer(reducer, initialState)
+
+  // provider methods
+  const setSource = source => {
+    // find and update source or
```
pzhine: I think we want "add or update source and set cache value"

packages/databyss-services/sources/SourcesProvider.js
```diff
@@ -0,0 +1,72 @@
...
+    if (state.cache[id]) {
+      return state.cache[id]
+    }
+    dispatch(fetchSource(id))
+    return null
+  }
+
+  return (
+    <SourceContext.Provider
+      value={[getSource, setSource, { cache: state.cache }]}
```
pzhine: This was my bad from the strategy. We can simplify to just `[getSource, setSource, state.cache]`

packages/databyss-services/sources/SourcesProvider.js
```diff
@@ -0,0 +1,72 @@
...
+import createReducer from '@databyss-org/services/lib/createReducer'
+import { SourceNotFoundError } from './SourceNotFoundError'
+import { initialState } from './reducer'
+import { fetchSource, saveSource } from './actions'
+
+const useReducer = createReducer()
+
+export const SourceContext = createContext()
+
+const Error = () => (
```
pzhine: Let's move this to `databyss-ui/components/Notify/ErrorFallback` and make it take a `message` prop

packages/databyss-services/sources/SourcesProvider.js
```diff
@@ -0,0 +1,72 @@
...
+
+  useEffect(
+    () => {
+      setSourcState(state.cache[sourceId])
+    },
+    [state.cache[sourceId]]
+  )
+
+  if (source instanceof SourceNotFoundError) {
+    return <Error />
```
pzhine: So this would become:
```js
if (source instanceOf Error) {
  // reset source in cache so we can retry
  removeCacheValue(sourceId)
}
if (source instanceOf ResourceNotFoundError) { 
  return <ErrorFallback message="Source not found" />
}
if (source instanceOf Error) {
  // throw to the source to trigger the "unexpected error" dialog
  throw source
}
```

packages/databyss-services/sources/SourcesProvider.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  return source ? <Wrapped source={source} {...others} /> : <Loading />
```
pzhine: Let's make a component `databyss-ui/components/Notify/LoadingFallback` that does the centering like `ErrorFallback`

packages/databyss-services/sources/actions.js
```diff
@@ -0,0 +1,87 @@
...
+        dispatch({
+          type: CACHE_SOURCE,
+          payload: { source, id: sourceFields._id },
+        })
+      })
+      .catch(() => {
+        dispatch({
+          type: CACHE_SOURCE,
+          payload: {
+            source: new SourceNotFoundError(
```
pzhine: At this point, we're not sure the error is a SourceNotFoundError.
In `request.js`, let's add an error class `ResourceNotFoundError` that we throw when the status is `404`. 
Then, we can just pass the error up:
```js
catch(err => {
   dispatch({
      payload: {
        source: err,
      }
   })
})
```

packages/databyss-ui/stories/Sources/sources.stories.js
```diff
@@ -0,0 +1,104 @@
...
+storiesOf('Accounts//Sources', module)
+  .addDecorator(ProviderDecorator)
+  .addDecorator(ViewportDecorator)
+  .add('Load and Save sources', () => (
+    <View>
+      <Box>
+        <SourcesDemo />
+      </Box>
+    </View>
+  ))
```
pzhine: We also need some error stories.

It would be helpful to see these in a more structured UI (like boxes with minWidth, minHeight) to see how the fallback components work.

packages/databyss-services/sources/SourceNotFoundError.js
```diff
@@ -0,0 +1,15 @@
+export class SourceNotFoundError extends Error {
```
pzhine: Let's make this a more generic `ResourceNotFoundError` thrown by `request.js` (see comment in `actions`)

packages/databyss-services/sources/SourcesProvider.js
```diff
@@ -0,0 +1,72 @@
+import React, { createContext, useContext, useState, useEffect } from 'react'
```
pzhine: File should be named `SourceProvider.js` for consistency

packages/databyss-services/sources/SourcesProvider.js
```diff
@@ -0,0 +1,72 @@
...
+
+export const useSourceContext = () => useContext(SourceContext)
+
+SourceProvider.defaultProps = {
+  initialState,
+}
+
+export const withSource = Wrapped => ({ sourceId, ...others }) => {
+  const [getSource, , state] = useSourceContext()
+  const [source, setSourcState] = useState(getSource(sourceId))
```
pzhine: Typo: `setSourceState`

packages/databyss-services/sources/SourcesProvider.js
```diff
@@ -0,0 +1,72 @@
...
+
+  useEffect(
```
pzhine: Move below error handling so that we don't re-render after a cache removal

packages/databyss-services/sources/SourcesProvider.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  const getSource = id => {
+    if (state.cache[id]) {
+      return state.cache[id]
+    }
+    dispatch(fetchSource(id))
+    return null
+  }
+
```
pzhine: Add a `removeCacheValue(id)` method

packages/databyss-services/sources/actions.js
```diff
@@ -0,0 +1,87 @@
+import * as sources from './'
+import { SourceNotFoundError } from './SourceNotFoundError'
+
+import { GET_SOURCE, SET_SOURCE, CACHE_SOURCE } from './constants'
```
pzhine: Let's standardize to `{ FETCH_SOURCE, SAVE_SOURCE, ...}`

packages/databyss-services/sources/SourcesProvider.js
msenejoa: `Rendered fewer hooks than expected. This may be caused by an accidental early return statement.`

it seems you cant have hooks after return statements

msenejoa:
fixes issue #127

packages/databyss-ui/stories/Sources/sources.stories.js
```diff
@@ -87,9 +108,9 @@ const Box = ({ children }) => (
...
 
 const ProviderDecorator = storyFn => (
-  <SourcesProvider initialState={initialState} reducer={reducer}>
+  <SourceProvider initialState={initialState} reducer={reducer}>
     {storyFn()}
-  </SourcesProvider>
+  </SourceProvider>
 )
 
 storiesOf('Accounts//Sources', module)
```
pzhine: Let's create a new top-level heading called `SERVICES` and put this `Sources` story in there, and also the "Login and Create Account Id".

packages/databyss-ui/stories/Errors/sourcesError.stories.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import ErrorFallback from '@databyss-org/ui/components/Notify/ErrorFallback'
+import { ViewportDecorator } from '../decorators'
+
+storiesOf('Notify//Sources', module)
```
pzhine: Doesn't need its own story file. We can put this in `databyss-ui/stories/Components/notify.stories.js`

packages/databyss-ui/stories/Loading/loading.stories.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
```
pzhine: Doesn't need its own story file. We can put this in `databyss-ui/stories/Components/notify.stories.js`

packages/databyss-api/src/models/Range.js
```diff
@@ -7,9 +7,12 @@ const Range = new mongoose.Schema({
   offset: {
     type: Number,
   },
-  marks: {
-    type: Array,
-  },
+  // example ['bold', 'italic', 'location']
+  marks: [
+    {
+      type: String,
```
pzhine: Should be [String], right?

PR #136, opened 2019-12-18
Michael/value list

msenejoa:
* mocked API in storybook
* navigation provider for modals
* editing atomic sources 
* full circle tests editing atomic sources connected to API


packages/databyss-ui/editor/slate/page/ContentEditable.js
```diff
@@ -272,8 +293,15 @@ const SlateContentEditable = forwardRef(
       if (hotKeys.isEsc(event)) {
         onNewBlockMenu(false, editor)
       }
+
+      if (isInlineSourceSelected(editor) && event.key === 'Enter') {
```
pzhine: Ok, I see why we need to pass `onEditSource` into `ContentEditable`. 
This event handler is higher level than the one inside `EditorInline`, so let's pass it into `renderInline` so all source edit actions are passing through the same higher-level function.
When we add topic editing, we'll want to abstract this handler (and related code) into something like `onEditEntity` so we're not repeating ourselves.

packages/databyss-ui/editor/slate/page/reducer.js
```diff
@@ -255,6 +258,41 @@ const deleteBlockById = id => (editor, value, next) => {
   next(editor, value)
 }
 
+/*
+updates all sources provided in the ID list
+*/
+const onUpdateSource = (source, idList) => (editor, value, next) => {
```
pzhine: Can we compute the idList here so we don't have to pass it in?

packages/databyss-ui/modules/SourcesValueList/SourceModal.js
```diff
@@ -0,0 +1,46 @@
...
+    // if (values) {
+    //   // updates in source provider
+    //   setSource(_values)
+    // }
+  }
+
+  const onDismiss = () => {
+    if (values) {
+      // updates in source provider
+      setSource(values)
```
pzhine: It looks like this only updates the source when the modal is dismissed, but we want them updating whenever a field blurs

packages/databyss-ui/modules/SourcesValueList/SourcesValueList.js
```diff
@@ -0,0 +1,101 @@
...
+import _ from 'lodash'
+
+const ControlList = ({ children, ...others }) => (
+  <List horizontalItemPadding="small" {...others}>
+    {children}
+  </List>
+)
+
+const SourcesValueList = ({ source, onValueChange }) => {
+  const [values, setValues] = useState(source)
```
pzhine: Too many intermediate states here. This change event should bubble up.

packages/databyss-ui/modules/SourcesValueList/SourceModal.js
```diff
@@ -0,0 +1,46 @@
...
+import { hideModal } from '@databyss-org/ui/components/Navigation/NavigationProvider/actions'
+import { useNavigationContext } from '@databyss-org/ui/components/Navigation/NavigationProvider/NavigationProvider'
+import SourcesValueList from './SourcesValueList'
+
+const SourceModal = ({ sourceId, visible, onUpdateSource }) => {
+  const { setSource } = useSourceContext()
+  const [values, setValues] = useState(null)
+  const [, dispatchNav] = useNavigationContext()
+
+  const onChange = _values => {
```
pzhine: Unnecessary function, we can just use `setValues` directly

packages/databyss-ui/primitives/Control/TextControl.js
```diff
@@ -50,6 +51,18 @@ const TextControl = ({
     },
     zIndex: active ? 2 : 'unset',
   }
+
+  useEffect(
```
pzhine: I think the "autofocus" feature should be moved to the `SourceModal` component, it can introduce weird side effects at this level of the render tree

packages/databyss-ui/theming/buttons.js
```diff
@@ -74,6 +74,14 @@ const buttonVariants = {
     paddingTop: pxUnits(6),
     borderRadius: pxUnits(0),
   },
+  editSource: {
```
pzhine: - Let's call this something more generic, like `editAtomicInline`
- Border radius should use global from theme
- Use theme space values instead of px values

packages/databyss-api/src/models/Source.js
```diff
@@ -3,43 +3,51 @@ const RichText = require('./RichText')
...
+      },
+    ],
+    entries: [
+      {
+        type: Schema.Types.ObjectId,
+        ref: 'entry',
+      },
+    ],
+  },
+  { minimize: false }
```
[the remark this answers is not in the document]
msenejoa: I removed this. [minimize](https://mongoosejs.com/docs/guide.html#minimize) is used in mongoose to save default empty values

packages/databyss-services/lib/request.js
```diff
@@ -37,7 +39,15 @@ function request(uri, options, responseIsJson) {
   const promise = fetch(uri, options)
     .then(checkStatus)
     .then(parseResponse(responseIsJson))
-
+    .catch(err => {
+      let _err = err
+      if (_err.message === "Cannot read property 'token' of null") {
+        _err = new NotAuthorizedError('not authorized')
+      } else if (_err.message === 'Failed to fetch') {
```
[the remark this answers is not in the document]
msenejoa: This error occurs when there is no token stored in local storage. The fetch API attempts to set the auth header to value `null`

packages/databyss-services/lib/request.js
```diff
@@ -37,7 +39,15 @@ function request(uri, options, responseIsJson) {
   const promise = fetch(uri, options)
     .then(checkStatus)
     .then(parseResponse(responseIsJson))
-
+    .catch(err => {
+      let _err = err
+      if (_err.message === "Cannot read property 'token' of null") {
```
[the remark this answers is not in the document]
msenejoa: If fetch is unable to make a connection, the response will not have a status code and is not accessible in the response. These errors must be caught in the catch portion of the fetch response. i found this [link ](https://github.com/github/fetch/issues/201#issuecomment-308213104) useful.

packages/databyss-services/pages/PageProvider.js
```diff
@@ -9,8 +12,18 @@ export const PageContext = createContext()
 const PageProvider = ({ children, initialState }) => {
   const [state, dispatch, stateRef] = useReducer(reducer, initialState)
 
+  const fetchPages = () => {
```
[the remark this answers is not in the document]
msenejoa: There would be no way of knowing in our storybook of what page to load since an `_id` is required to look up a page in the API. This is currently used for the initial mount of `Page` under our `Services` in storybook. The `EditorLoader` component is a consumer of `withPages`.

packages/databyss-services/sources/SourceProvider.js
```diff
@@ -49,20 +71,25 @@ SourceProvider.defaultProps = {
...
   if (source instanceof Error) {
-    // throw to the source to trigger the "unexpected error" dialog
-    throw source
+    return <ErrorFallback error={source} />
   }
 
   return source ? <Wrapped source={source} {...others} /> : <Loading />
 }
 
+export const withSources = Wrapped => ({ sourceList, ...others }) => {
```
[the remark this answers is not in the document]
msenejoa: as mentioned above, it's currently used in our storybook for the initial load of `SERVICES||Page`.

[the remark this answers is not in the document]
msenejoa: removed

packages/databyss-services/sources/actions.js
```diff
@@ -39,25 +46,31 @@ export function saveSource(sourceFields) {
...
         dispatch({
           type: CACHE_SOURCE,
-          payload: { source, id: sourceFields._id },
+          payload: { source: sourceFields, id: sourceFields._id },
         })
       })
-      .catch(() => {
+      // check for type of error
+      // send correct error class
+      .catch(err => {
```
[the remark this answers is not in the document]
msenejoa: removed

packages/databyss-services/sources/actions.js
```diff
@@ -74,6 +87,22 @@ export function removeSourceFromCache(id) {
   }
 }
 
+export function fetchAllSources() {
```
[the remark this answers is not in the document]
msenejoa: This is used in the `EditorPage.js` `useEffect` hook when looking up if a source exists. or if a new one has been added and must be cached. An alternative to this would be to load all the sources a page currently contains on the initial mount of that specific page. Either way, we must have all the sources cached if we want to compare our `editorState` to the `sourceProvider`.

packages/databyss-services/sources/constants.js
```diff
@@ -1,4 +1,5 @@
 export const FETCH_SOURCE = 'FETCH_SOURCE'
+export const GET_ALL_SOURCES = 'GET_ALL_SOURCES'
```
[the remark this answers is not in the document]
msenejoa: explained in previous comment.

packages/databyss-ui/editor/EditorInline.js
```diff
@@ -1,20 +1,97 @@
...
     return (
-      <Styled {...styleProps} {...others} ref={ref}>
+      <Span
+        {...others}
+        //  onClick={onClick}
+        onMouseDown={onClick}
+        ref={ref}
+        id="test"
+        borderRadius={5}
+        border="1px"
```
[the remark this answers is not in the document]
msenejoa: removed

packages/databyss-ui/editor/EditorInline.js
```diff
@@ -1,20 +1,97 @@
...
+    }
+
     return (
-      <Styled {...styleProps} {...others} ref={ref}>
+      <Span
+        {...others}
+        //  onClick={onClick}
+        onMouseDown={onClick}
+        ref={ref}
+        id="test"
```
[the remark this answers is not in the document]
msenejoa: removed. It was used for debugging.

packages/databyss-ui/editor/EditorPage.js
```diff
@@ -14,12 +21,46 @@ import {
...
+  checks to see if new source has been added
+  adds the new source to the source provider
+  */
+  useEffect(
+    () => {
+      const _sourceDictionary = getSources()
+      const _sources = Object.keys(sources)
+      _sources.forEach(_refId => {
+        if (!_sourceDictionary[_refId]) {
+          // TODO: NEED TO GET SOURCE FIRST
```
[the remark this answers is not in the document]
msenejoa: Another alternative would be to just cache all the sources currently in our `editorState` on the initial page load, which would make them readily accessible.

packages/databyss-ui/editor/EditorPage.js
```diff
@@ -80,6 +121,27 @@ const EditorPage = ({ children }) => {
     dispatchEditor(newBlockMenu(bool, { value }))
   }
 
+  const [, dispatchNav] = useNavigationContext()
+
+  const onEditSource = (refId, blocks, { value }) => {
```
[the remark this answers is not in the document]
msenejoa: This is for when the user presses `enter` while the source block is highlighted. All of the key captures occurs in the `contentEditable` component.

packages/databyss-ui/editor/slate/page/ContentEditable.js
msenejoa: I attempted this and got suck on passing extra properties to `renderInline`. Lets discuss this next time we talk.

packages/databyss-ui/editor/slate/page/reducer.js
msenejoa: the `idList` can be computed in the slate reducer but we will still need to pass the `blocks` property since the slate reducer does not have access to our state values. This can be computed with the editor state when the `refId` is attached to the `data` attribute of each node.

packages/databyss-ui/modules/SourcesValueList/SourceModal.js
msenejoa: Updating the sourcesProvider `onBlur` causes the value list to re-render and loose focus. we can discuss this on our next chat

packages/databyss-ui/modules/SourcesValueList/SourcesValueList.js
msenejoa: Since `SourceValueList` is a consumer of `withSource` this requires the parent component to use `setSource` in order to update the source, this causes a re-render which looses focus on every key change.

packages/databyss-ui/modules/SourcesValueList/SourceModal.js
msenejoa: ill keep this function until we discuss the best way to update the source state without causing a re-render.

packages/databyss-ui/editor/EditorInline.js
```diff
@@ -1,20 +1,97 @@
...
+      [node, editor]
+    )
+
+    const onUpdateSource = source => {
+      // return a list of blocks containing the source that will be updated
+      if (source) {
+        const _idList = Object.keys(blocks).filter(
+          block => blocks[block].refId === source._id
+        )
+        dispatchEditor(updateSource(source, _idList, { value: editor.value }))
```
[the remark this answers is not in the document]
msenejoa: The editableReducer does not have access to our internal `state`

packages/databyss-api/src/models/Author.js
```diff
@@ -1,21 +1,7 @@
...
-//     type: String,
-//     default: '',
-//   },
-// })
-
 const mongoose = require('mongoose')
 
 const Author = new mongoose.Schema({
-  textValue: { type: String, default: 's' },
+  textValue: { type: String, default: '' },
```
pzhine: Shouldn't this have `firstName` and `lastName`?

packages/databyss-api/src/models/Source.js
```diff
@@ -3,51 +3,47 @@ const RichText = require('./RichText')
...
+  account: {
+    type: Schema.Types.ObjectId,
+    ref: 'account',
+  },
+  resource: {
+    type: String,
+  },
+  text: RichText,
+  citations: { type: Array, default: [{ textValue: '', ranges: [] }] },
+  authors: {
```
pzhine: Let's see if we can declare this as type `[Author]` and have the default be `[Author()]`

packages/databyss-services/lib/request.js
```diff
@@ -4,16 +4,14 @@ import { ResourceNotFoundError } from './ResourceNotFoundError'
...
-
 function checkStatus(response) {
   if (response.status >= 200 && response.status < 300) {
     return response
   }
   if (response.status === 401) {
     deleteAuthToken()
-    window.location = '/login'
-    throw new UnauthorizedError('Unauthorized')
+    // window.location = '/login'
```
pzhine: Add a TODO here to move the redirect upstream

packages/databyss-services/pages/actions.js
```diff
@@ -50,48 +47,37 @@ export function savePage(state) {
...
-    //     type: PAGE_LOADED,
-    //     payload: res,
-    //   })
-    // }
-    // )
+    services
+      .getAllPages()
+      .then(res => {
+        dispatch({
+          type: PAGES_LOADED,
```
pzhine: Why do we need `getAllPages()`?

packages/databyss-services/sources/SourceProvider.js
```diff
@@ -34,8 +41,10 @@ const SourceProvider = ({ children, initialState, reducer }) => {
 
   const getAllSources = () => {
     dispatch(fetchAllSources())
+  }
 
-    //  console.log('here')
+  const getSourcesforPage = id => {
```
pzhine: This needs to use the cache as well.

packages/databyss-services/sources/SourceProvider.js
```diff
@@ -44,9 +53,7 @@ const SourceProvider = ({ children, initialState, reducer }) => {
     }
   }
 
-  const getSources = () => {
-    return state.cache
-  }
+  const getSources = () => state.cache
```
pzhine: This method is confusing. Do we need it?

packages/databyss-services/sources/SourceProvider.js
```diff
@@ -56,6 +63,7 @@ const SourceProvider = ({ children, initialState, reducer }) => {
         removeCacheValue,
         getSources,
         getAllSources,
+        getSourcesforPage,
```
pzhine: This should match the naming of the fetch, so `getPageSources`

packages/databyss-services/sources/actions.js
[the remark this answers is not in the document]
pzhine: No we can remove this and use `fetchPageSources`, right?

packages/databyss-ui/editor/EditorPage.js
```diff
@@ -26,11 +22,11 @@ import {
 
 import { isBlockEmpty, isEmptyAndAtomic } from './slate/slateUtils'
 
-const EditorPage = ({ children }) => {
+const EditorPage = ({ children, autoFocus }) => {
   const [editorState, dispatchEditor] = useEditorContext()
-  const { getSource, setSource, getSources } = useSourceContext()
+  const { setSource, getSources } = useSourceContext()
```
pzhine: We should use `getPageSources` here, right?

packages/databyss-ui/editor/EditorPage.js
```diff
@@ -126,18 +118,17 @@ const EditorPage = ({ children }) => {
...
     }
     dispatchNav(
-      showModal('SOURCE', {
-        sourceId: refId,
-        onUpdateSource,
+      showModal({
+        component: 'SOURCE',
+        props: {
+          sourceId: refId,
+          onUpdateSource,
```
pzhine: Can the `SourceModal` consume the `SourceContext` directly to perform the update, so that we don't have to pass the function in? Then we could just pass in `value`. Also, it doesn't seem like we use `blocks`. Can we remove it?

packages/databyss-services/sources/SourceProvider.js
pzhine: 1. Resolve sourceIds for the page
2. Check in cache for some/none/all sourceIds
3. Build list of missing sourceIds
4. Fetch missing sourceIds

packages/databyss-services/sources/SourceProvider.js
pzhine: Only used in `EditorPage`, but we want to use `getPageSources` instead there. 
Also used in `fetchAllSources`, which we can remove

packages/databyss-ui/editor/EditorPage.js
pzhine: This is because `SourceModal` lives outside of `EditorProvider`

packages/databyss-services/sources/SourceProvider.js
msenejoa: This method is needed to check the state of the cache for composing a list of sources not already in the cache. Ill change this to `getCache`

msenejoa: i changed this from `getSources()` to `state`

PR #146, opened 2020-01-08
Michael/page provider

msenejoa:
* refactor Page Provider
* added new source buffer to Editor reducer
* refactor storybook demo


packages/databyss-services/pages/PageProvider.js
```diff
@@ -13,18 +13,30 @@ export const PageContext = createContext()
...
-  const fetchPages = () => {
-    if (state.pages) {
-      return state.pages
+  const getPages = () => {
+    if (state.headerCache) {
+      return state.headerCache
     }
-    if (!state.isPagesLoading) {
-      dispatch(getPages())
+    if (!state.isLoading) {
```
msenejoa: if this property is removed. Editor gets in a loop of fetching pages

pzhine: Ah, true.
So right now our cache values have 2 states based on type: `null` (unfetched) and `Page` (fetched)
What if we have 3 states: `null` (unfetched), `Promise` (fetching) and `Page` (fetched)

Note more explicit naming of pages -> pageHeaders throughout, and normalizing actions to `FETCH_` and `CACHE_` (actions should be command-like).

```js
// PageProvider.js
  const getPageHeaders = () => {
    if (state.headerCache) {
      return state.headerCache
    }
    if (!state.headerCache instanceof Promise) {
      dispatch(fetchPageHeaders())
    }
    return null
  }
// ...
export const PagesLoader = ({ children }) => {
  const { getPageHeaders } = usePageContext()
  const pages = getPageHeaders()

  if (pages instanceof Error) {
    return <ErrorFallback error={pages} />
  }

  if (typeof children !== 'function') {
    throw new Error('Child must be a function')
  }

  if (!pages || pages instanceof Promise) {
    return <Loading />
  }

  return children(pages)
}

// actions.js
export function fetchPageHeaders() {
  return dispatch => {
    const promise = services
      .getAllPages()
      .then(res => {
        dispatch({
          type: CACHE_PAGE_HEADERS,
          payload: res,
        })
      })
      .catch(e => {
        dispatch({
          type: CACHE_PAGE_HEADERS,
          payload: e,
        })
      })
    dispatch({
      type: FETCH_PAGE_HEADERS,
      payload: { promise },
    })
  }
}

// reducer.js
    case FETCH_PAGE_HEADERS: {
      return {
        ...state,
        headerCache: action.payload.promise,
      }
    }
```

packages/databyss-ui/editor/AutoSave.js
```diff
@@ -1,31 +1,32 @@
...
       }
       if (hasModal) {
-        clearInterval(refreshId)
-        setRefreshId(null)
+        clearInterval(_timer)
+        _timer = null
+      }
+      // on unmount clear autosave
+      return () => {
+        clearInterval(_timer)
```
pzhine: Might as well do one last `savePage` here, right?

packages/databyss-ui/editor/EditorPage.js
```diff
@@ -18,36 +18,35 @@ import {
   deleteBlocks,
   newBlockMenu,
   updateSource,
+  removeSourceFromQue,
```
pzhine: Either spell it `Queue` or use just `Q`

packages/databyss-ui/editor/state/page/actions.js
```diff
@@ -16,6 +16,7 @@ import {
   SHOW_FORMAT_MENU,
   SHOW_NEW_BLOCK_MENU,
   UPDATE_SOURCE,
+  DEQUEUE_NEW_SOURCES,
```
pzhine: The plural is misleading. I was imagining it taking an array, but the singular works too. Make the action name singular.

packages/databyss-ui/editor/state/page/reducer.js
```diff
@@ -128,8 +129,9 @@ const setBlockType = (state, type, _id) => {
 
   switch (type) {
     case 'SOURCE':
-      nextState.sources[nextRefId] = { _id: nextRefId, textValue, ranges }
-
+      const _source = { _id: nextRefId, textValue, ranges }
+      nextState.sources[nextRefId] = _source
+      nextState.newSources = [_source]
```
pzhine: If there's still a source in the q, this will clobber it. I think we want `nextState.newSources.push(_source)`

packages/databyss-ui/editor/EditorPage.js
msenejoa: fixed

Issue #142, opened 2020-01-06
Offset atomic block to align text


pzhine:
**Observed behavior**
1. Go to Editor demo
2. Create a long new source that spans multiple lines
3. Text of first line of source is indented slightly compared to 2nd line (or entry text)

**Expected behavior**
Text of source should be aligned flush with the rest of the text.

**Screenshots**
<img width="483" alt="Screenshot 2020-01-07 00 22 30" src="https://user-images.githubusercontent.com/2692844/71856515-7f060880-30e4-11ea-807e-c291f8a4e565.png">

PR #149, opened 2020-01-16
Michael/copypaste

msenejoa:
* copy and paste feature added
* `refId` is now stored inside of the the block `data` parameter


.storybook/server.cert
```diff
@@ -0,0 +1,22 @@
+-----BEGIN CERTIFICATE-----
```
pzhine: I think we can remove the cert and key files.

packages/databyss-ui/editor/slate/page/ContentEditable.js
```diff
@@ -26,6 +26,7 @@ import {
   noAtomicInSelection,
   getSelectedBlocks,
   isInlineSourceSelected,
+  NewEditor,
```
pzhine: Why does this need to be passed in as a dep?

packages/databyss-ui/editor/slate/page/ContentEditable.js
```diff
@@ -517,16 +519,51 @@ const SlateContentEditable = forwardRef(
...
+                return true
+              }
+              return false
+            })
+            const _node = editor.value.document.getNode(_id).toJSON()
+            const _editor = NewEditor()
+            _editor.insertFragment(_frag)
+            const _nodeList = _editor.value.document.nodes.map(n => n.key)
+            _editor.replaceNodeByKey(_nodeList.get(i), _node)
+            _frag = _editor.value.document
```
pzhine: If `_frag` isn't saved or accumulated at the end of this loop, it's lost. Based on the logic, it seems we can handle only one block where type is SOURCE. But there might be more, right?

It's very hard to figure out what this loop is doing because it has so many side effects. Can you explain what is happening by describing the state before and after this code runs?

pzhine:
Bugs:
- [ ]  Pasting a SOURCE that no longer exists should work
- [ ] CUT fails with more than one line selected

pzhine:
Bugs:
- [ ] empty lines before and/or after the selected line should be included in the paste
- [ ] when selecting an empty line before an atomic, the selection is extended to cover the atomic. (it shouldn't)
- [ ] hide block menu launcher (+) if anything is selected

pzhine:
Bug:
- [ ] Cut doesn't go to clipboard when > 1 line is selected

packages/databyss-ui/editor/slate/page/ContentEditable.js
msenejoa: `blockList` and `frag` return a cumulative value.

pzhine:
- [ ] if `transfer.text` is null, don't try to paste anything
- [ ] disable "drag and drop" in Slate

msenejoa:
* added cut
* added `type` to `data` to block node
* extends selections to include `atomic blocks`

packages/databyss-ui/editor/slate/clipboard.js
```diff
@@ -160,3 +161,218 @@ export const isFragmentFullBlock = (fragment, document) => {
...
   return false
 }
+
+/*
+        looks up the refId of the fragment and replaces it with an updated value, paste blocks can become stale when copying and pasting
+
+        this function takes a blockList, fragment, value and currentState and returns updated { blockList, fragment } 
+        */
+
+export const updateClipboardRefs = ({ blockList, fragment, sourceCache }) => {
```
pzhine: Please annotate parameter as a list, like we do in the API package. Describe the data types.
```
// @blockList    array of slate blocks
// @fragment    Slate value
...
```

packages/databyss-ui/editor/slate/clipboard.js
```diff
@@ -160,3 +161,218 @@ export const isFragmentFullBlock = (fragment, document) => {
...
+  const _slateBlockList = blockList
```
pzhine: This is unnecessary, just use `blockList`

packages/databyss-ui/editor/slate/clipboard.js
pzhine: Also, the param list in the comment doesn't match the param list in the code...

packages/databyss-ui/editor/slate/clipboard.js
```diff
@@ -160,3 +161,218 @@ export const isFragmentFullBlock = (fragment, document) => {
...
+    if (_frag.nodes.size > 1) {
+      _frag = trimFragment(_frag)
+    }
+
+    // get list of refId and Id of fragment to paste,
+    // this list is used to keep slate and state in sync
+    let _blockList = blocksToState(_frag.nodes)
+
+    /*
+        looks up the refId of the fragment and replaces it with an updated value, paste blocks can become stale when copying and pasting
```
pzhine: No need to repeat the comment when the function is called

packages/databyss-ui/editor/slate/clipboard.js
```diff
@@ -160,3 +161,218 @@ export const isFragmentFullBlock = (fragment, document) => {
...
+      blockList: _blockList,
+      fragment: _frag,
+      offset: _offset,
+      beforeBlockId: _beforeBlockId,
+      beforeBlockRef: _beforeBlockRef,
+      afterBlockId: _afterBlockId,
+      afterBlockRef: _afterBlockRef,
+    }
+    return _pasteData
+    // onPasteAction(_pasteData, editor)
```
pzhine: Cleanup comments

pzhine:
- [ ] remove space between atomic and hanging inline (see screenshot)
- [ ] add negative margin to atomic blocks so that text aligns (#142)

pzhine:
- [ ] CUT bug
ezgif-1-602e3cede713

pzhine:
- [ ] BACKSPACE bug
ezgif-1-854edcec6ebf

packages/databyss-ui/editor/EditorInline.js
```diff
@@ -24,6 +24,8 @@ const EditorInline = React.forwardRef(
         ref={ref}
         borderRadius={5}
         p="tiny"
+        pr="0"
+        ml="-3px"
```
pzhine: use `pxValue` for cross-platform
better yet, add a new size called `negativeTiny` so that they always match

packages/databyss-services/sources/actions.js
```diff
@@ -12,6 +12,7 @@ import {
 } from './constants'
 
 export function fetchSource(id) {
+  console.log(id)
```
pzhine: cleanup comment

PR #151, opened 2020-01-22
Michael/topics provider

msenejoa:
* topics provider
* topics modal
* cypress tests


packages/databyss-services/topics/actions.js
```diff
@@ -0,0 +1,45 @@
...
+}
+
+export function saveTopic(topicFields) {
+  return async dispatch => {
+    dispatch({
+      type: SAVE_TOPIC,
+      payload: { topic: topicFields, id: topicFields._id },
+    })
+    topics.setTopic(topicFields).then(() => {
+      dispatch({
```
pzhine: We don't need a `CACHE_TOPIC` here because `SAVE_TOPIC` already updates the cache.

packages/databyss-ui/editor/EditorPage.js
```diff
@@ -25,29 +27,46 @@ import { isBlockEmpty, isEmptyAndAtomic } from './slate/slateUtils'
...
-        if (newSources.length > 0) {
-          newSources.forEach(s => {
-            const _source = {
-              _id: s._id,
-              text: { textValue: s.textValue, ranges: s.ranges },
-            }
-            setSource(_source)
-            dispatchEditor(removeSourceFromQueue(s._id))
-          })
+      if (editableState) {
```
pzhine: Change to guard pattern to avoid nesting levels:
```
if (!editableState) {
  return
}
```

packages/databyss-ui/editor/EditorPage.js
```diff
@@ -25,29 +27,46 @@ import { isBlockEmpty, isEmptyAndAtomic } from './slate/slateUtils'
...
+        if (newSources) {
```
pzhine: Combine these to `newAtomics` as discussed. 
Also, rather than nested `if`s, combine to `if (newAtomics && newAtomics.length > 0)`

packages/databyss-ui/editor/slate/page/reducer.js
```diff
@@ -228,18 +228,18 @@ const deleteBlockById = id => (editor, value, next) => {
 }
 
 /*
-updates all sources provided in the ID list
+updates all atomics provided in the ID list
 */
-const onUpdateSource = (source, blocks) => (editor, value, next) => {
+const onUpdateAtomic = (atomic, blocks, type) => (editor, value, next) => {
```
pzhine: For more than 2 params, use object dereferencing

packages/databyss-ui/editor/state/page/actions.js
```diff
@@ -180,12 +181,13 @@ export function newBlockMenu(bool, editableState) {
   }
 }
 
-export function updateSource(source, editableState) {
+export function updateAtomic(atomic, type, editableState) {
```
pzhine: Use object deref for params here too

packages/databyss-ui/modules/SourcesValueList/SourceModal.js
```diff
@@ -21,7 +21,7 @@ const ControlList = ({ children, ...others }) => (
   </List>
 )
 
-const SourceModal = ({ sourceId, visible, onUpdateSource, id }) => {
+const SourceModal = ({ atomicId, visible, onUpdate, id }) => {
```
pzhine: Let's use `refId` here, for consistency

PR #170, opened 2020-02-07
set block ref in use effect

msenejoa:
* sets the block ref in a use effect hook


pzhine:
Closing because this fixes a bug exposed only by the perf testing in #169 right now. We will adapt #169 to work with the new `@databyss-org/editor` package, which shouldn't have this issue.

PR #180, opened 2020-03-12
Michael/pages sidebar

msenejoa:
- Sidebar with pages


pzhine:
- [ ] upon new account creation, default page should be given "untitled" as name
- [ ] remove arrows from page names
- [ ] collapsed view page icon is shortcut to expanded sidebar in pages view

PR #191, opened 2020-03-27
Michael/search

msenejoa:
* added search feature


packages/databyss-ui/modules/Sidebar/Sidebar.js
```diff
@@ -133,6 +72,9 @@ const Sidebar = () => {
         </PagesLoader>
       )
     }
+    if (menuItem === 'search') {
```
pzhine: Let's move `menuItem` out of local state to `NavigationProvider`, and let's use the same route convention as the content area. So instead of `setMenuItem('search')` would do something like `NavigationProvider.navigateSidebar('/search')`. Then we won't have to pass 'setMenuItem` and `onClick` down through props, but instead can just consume `NavigationProvider` in child components where we need them.

packages/databyss-ui/modules/Sidebar/Sidebar.js
```diff
@@ -133,6 +72,9 @@ const Sidebar = () => {
...
+      return null
+    }
     return SidebarList({
```
pzhine: We should migrate `SidebarList` to `routes/Pages` to make the handling consistent.

packages/databyss-ui/components/SearchContent/SearchContent.js
```diff
@@ -0,0 +1,94 @@
...
+  ScrollView,
+} from '@databyss-org/ui/primitives'
+import PageSvg from '@databyss-org/ui/assets/page.svg'
+
+const SearchContent = () => {
+  const { getTokensFromPath, navigate } = useNavigationContext()
+
+  /*
+    todo: gets two different search queries
+  */
```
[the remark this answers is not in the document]
msenejoa: old comment before the cache search was implemented

packages/databyss-ui/editor/EditorBlock.js
```diff
@@ -120,10 +123,12 @@ export const EditorBlock = ({ children, node }) => {
...
     </Grid>
   ) : (
-    <View mb="tiny">{_children}</View>
+    <View mb="small">{_children}</View>
   )
 }
 
 export const renderBlock = ({ node, children }) => (
-  <EditorBlock node={node}>{children}</EditorBlock>
+  <Element name={node.key} id={node.key} className="element">
```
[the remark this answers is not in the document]
msenejoa: implemented a `refDictionary`

msenejoa:
resolves #194

packages/databyss-ui/editor/EditorBlock.js
```diff
@@ -86,6 +102,7 @@ export const EditorBlock = ({ children, node }) => {
 
   const _children = (
     <View
+      ref={blockEl}
```
pzhine: This can be cleaned up by using the callback form of `ref`. This avoids the need for lines 58-70.

```jsx
<View 
  ref={ref => pageContext && pageContext.registerBlockRef(node.key, ref)}
>
```

packages/databyss-services/pages/PageProvider.tsx
```diff
@@ -60,8 +61,30 @@ const PageProvider: React.FunctionComponent<PropsType> = ({
     return null
   }
 
+  const registerBlockRef = (id: string, ref: React.Ref<HTMLInputElement>) => {
+    dispatch(registerRef(id, ref))
```
pzhine: I think we should maintain the refDict as a "class variable" (ref) in PageProvider (i.e. outside of the reducer state), as they are low-level UI references and don't need dispatch/listener features. Let's discuss though.

packages/databyss-ui/modules/Sidebar/Sidebar.js
```diff
@@ -116,23 +108,24 @@ const Sidebar = () => {
...
+          <Footer />
         </View>
       </View>
     </View>
   ) : (
-    <SidebarCollapsed
-      onToggleMenuOpen={onToggleMenuOpen}
-      setMenuItem={setMenuItem}
-    />
+    <SidebarCollapsed onToggleMenuOpen={onToggleMenuOpen} />
```
pzhine: Move `menuOpen` state to `NavigationContext.state.sidebarExpanded` so we don't need to pass props

packages/databyss-ui/modules/Sidebar/routes/SidebarList.js
```diff
@@ -45,31 +45,32 @@ const menuSvgs = type =>
...
 
-const SidebarList = ({
-  menuItems = defaultMenu,
-  menuItem,
-  menuOpen,
-  //   onToggleMenuOpen,
-  onItemClick,
-}) => {
-  const { getTokensFromPath } = useNavigationContext()
+const SidebarList = ({ menuItems = defaultMenu, menuOpen }) => {
```
pzhine: `SidebarList` is really the root of the Sidebar routes, so let's call it `routes/Root.js`. Why does it need a `menuItems` param?

pzhine: The rendering parts of `SidebarList` should be moved to `components/Sidebar`. Anything specific to loading and compiling the Pages list should be moved to `routes/Pages`. And `defaultMenu` should be moved to `routes/Root`

PR #204, opened 2020-04-13
Michael/citations

msenejoa:
* implements citations search using google api


pzhine:
If only last name is specified, don't put an empty first name in the "Name" field. 

This should just be "Plato. Republic (2008)"
Screenshot 2020-04-14 15 13 15

pzhine:
- [ ] Show title and subtitle in results (below should be "Intertwingled: The Work and Influence of Ted Nelson")
- [ ] One book shouldn't be listed more than once. If there are co-authors, they should be coalesced, e.g. "Douglas R. Dechow, Daniele C. Struppa"
Screenshot 2020-04-14 15 19 51

pzhine:
- [ ] Only show results if all search terms appear in either author or title (including co-authors and subtitles). In other words, each word in the query must appear at least once in one of these fields.

- [ ] Multi-line titles should still show year at the end of the title, not on the next line
Screenshot 2020-04-14 15 46 25

pzhine:
- [ ] ESC key should hide suggestions tooltip and modal
- [ ] Lock page scroll when suggestions modal is visible
Screenshot 2020-04-14 15 54 15

packages/databyss-ui/lib/useEventListener.js
```diff
@@ -0,0 +1,39 @@
+import { useRef, useEffect } from 'react'
```
pzhine, line 1: Let's credit the source of this snippet (include url in comments)

PR #211, opened 2020-04-20
Michael/format menu

msenejoa:
* implement format hovering menu to new editor


packages/databyss-editor/components/ContentEditable.js
```diff
@@ -174,6 +193,29 @@ const ContentEditable = () => {
...
+    // set_node is called on format change transforms
+    if (editor.operations.find(op => op.type === 'set_node')) {
+      // get indexies of selected nodes
+      const _blocksChanged = getSelectionIndicies(selection)
+
+      _blocksChanged.forEach(idx => {
+        // node should not be updated if a toggle mark occured
+        if (Node.string(value[idx])) {
+          // update target node
+          setContent({
```
pzhine, line 206: We need to change `setContent` to take an array of operations rather than calling it multiple times. The reason is that the dispatches seem to get queued up and then all called in sequence before the next `render`. That means the `operations` array and `preventDefault` flag get ignored in all but the final `setContent`, allowing the behavior shown below:
ezgif com-video-to-gif (3)

PR #221, opened 2020-04-27
Michael/citation menu

msenejoa:
* citations drop down menu


packages/databyss-editor/lib/markup.js
```diff
@@ -25,7 +24,12 @@ export const applyRange = (editor, range) => {
   // and correct the offset with an additional move, below.
   Transforms.move(editor, { distance: 1, edge: 'anchor', reverse: true })
 
-  toggleMark(editor, range.mark)
+  // apply marks array
+  if (range.marks) {
```
pzhine, line 28: I think we should keep the `range` element to a single mark, because that allows metadata to be added specific to that mark, and keeps our data structures from getting too deeply nested.

packages/databyss-editor/state/reducer.js
```diff
@@ -285,6 +288,13 @@ export default (state, action) =>
           draft.newEntities = _newEntities
         }
       }
+      // if next selection doesnt exist, replace selection with origin
+      if (!entityForBlockIndex(draft, draft.selection.focus.index)) {
```
pzhine, line 292: Should we do the same check for `selection.anchor.index` just in case?

pzhine:
- [ ] Fix topic and source regressions
  - Topic regression: type "# topic" on a line and press ENTER
  - Source regression: type "@ source" on a line, press ESC, then ENTER

PR #227, opened 2020-05-01
Michael/format bug fix

msenejoa:
* corrects range offset when atomic block gets `baked`
* removes location from atomic blocks
* selection bug fix


pzhine:
While testing this manually, we found some cases that break. As there are a lot of use cases to fix here, I think it's better we wait until we have some tests in `next`, rebase, and go from there.

packages/databyss-editor/components/ContentEditable.js
```diff
@@ -306,8 +306,10 @@ const ContentEditable = () => {
     // correct path (pointing within the mark leaf) but the anchor gets the parent
     // path. The fix for this is to overshoot the anchor by 1
     // and then correct the offset with an additional move.
-    Transforms.move(editor, { distance: 1, edge: 'anchor' })
-    Transforms.move(editor, { distance: 1, edge: 'anchor', reverse: true })
+    if (editor.selection && Range.isExpanded(editor.selection)) {
```
pzhine, line 309: Please explain why this condition is necessary in a comment

packages/databyss-editor/state/reducer.js
```diff
@@ -34,7 +36,7 @@ export const bakeAtomicBlock = ({ state, draft, index }) => {
       const _nextEntity = {
         text: {
           textValue: _entity.text.textValue.substring(1).trim(),
-          ranges: _entity.text.ranges,
+          ranges: removeLocationMark(offsetRanges(_entity.text.ranges, 1)),
```
pzhine, line 39: Please explain the purpose of `offsetRanges`

packages/databyss-editor/state/util.ts
```diff
@@ -40,3 +40,9 @@ export const getIndeciesForRefId = (state: any, refId: string) => {
 
   return matches
 }
+
+export const offsetRanges = (ranges: any, offset: any) =>
```
pzhine, line 44: Looks like you're ignoring the `offset` param

packages/databyss-editor/state/util.ts
```diff
@@ -40,3 +40,9 @@ export const getIndeciesForRefId = (state: any, refId: string) => {
...
+  ranges.map(r => ({ ...r, offset: r.offset - 1 }))
+
+export const removeLocationMark = (ranges: any) =>
```
pzhine, line 47: Don't we have a data type for `range`?

packages/databyss-editor/state/util.ts
pzhine: Also `offset` can have type `number`

packages/databyss-editor/interfaces/ranges.ts
```diff
@@ -0,0 +1,2 @@
+import { Range } from './'
+export type Ranges = Array<Range>
```
pzhine, line 2: Can we just use `Array<Range>` or `Range[]` in the function signature? Usually you don't need a separate type for the array version of a type...

packages/databyss-editor/state/util.ts
```diff
@@ -41,8 +41,40 @@ export const getIndeciesForRefId = (state: any, refId: string) => {
...
+    if (!offset) {
+      length = length - 1
+    } else {
+      offset = offset - _offset
+    }
+    return { ...r, length, offset }
+  })
 
-export const removeLocationMark = (ranges: any) =>
+export const removeLocationMark = (ranges: Ranges) =>
```
pzhine, line 58: See comment about `Ranges` type...

PR #236, opened 2020-05-13
paul/icon-styling

pzhine:
refactor Icons to record fill and stroke preferences for each icon


pzhine:
closing because we have to use fills rather than strokes for consistent rendering across browsers

PR #239, opened 2020-05-14
Michael/new editor

msenejoa:
* integrates new slate to app


packages/databyss-editor/components/Element.js
```diff
@@ -134,6 +136,20 @@ const Element = ({ attributes, children, element }) => {
...
 
+  const registerBlockRef = (ref = attributes.ref.current) => {
+    const _index = ReactEditor.findPath(editor, element)[0]
+    const _refId = editorContext.state.blocks[_index]._id
+    if (!pageContext.getBlockRef(_refId)) {
+      pageContext.registerBlockRef(_refId, ref)
+    }
+  }
+
+  useEffect(() => {
```
pzhine, line 147: This will fire after every render. Why is this necessary is we register the ref in the `View` component below?

PR #244, opened 2020-05-19
Michael/editor refactor single

msenejoa:
* remove old slate and add new single line slate rich text component


pzhine:
Also, let's remove the old editor components and all references to slate 0.4x

PR #264, opened 2020-06-09
Michael/bugs

### LINKED ISSUE #257: Error when Internet Goes Off
When the Internet is off or goes off in the middle of using the editor, the error message pops up. Sometimes this can even mess up the page entirely.

We should combat this internally.

### LINKED ISSUE #259: Search keeps deleted blocks
I believe the search is sending you to blocks that had the search term in it, even though they are deleted now.

msenejoa:
* fixes #259 
* fixes #257
* fix page title not saving when switching pages
* space between authors first and last name
* refine citations search filtering
* add tabs to the editor
* added em dash
* if invalid selection is saved on database, reset selection to origin
* added a patch buffer for offline mode
* offline modal that blocks all interactions 
* offline tests


packages/databyss-services/pages/reducer.js
```diff
@@ -36,12 +37,23 @@ export default (state, action) => {
         ..._state,
       }
     }
+    case REMOVE_PAGE_FROM_CACHE: {
+      const _cache = state.cache
```
pzhine, line 41: Here we need to do a deepClone because we're mutating the state with the `delete` statement. This will change once we move to `immer`

pzhine:
- [ ] Test(s) for fixes for #259 and #257 
- [ ] Test for round-trip integrity after switching pages (should break before this merge)

PR #271, opened 2020-06-19
Michael/optimize search

msenejoa:
* searches to include only exact match of word
* search for multiple words
* move selection to focus index in editor when word is searched
* moved search term from `Search` to `EntryProvider` state
* highlight search terms within the editor


packages/databyss-ui/components/SearchContent/SearchContent.js
```diff
@@ -60,8 +60,9 @@ const SearchContent = () => {
               <View p="small" ml="small">
                 <Text>
                   <Highlighter
-                    searchWords={[query]}
-                    autoEscape
+                    searchWords={query
```
pzhine, line 63: Are we still using `Highlighter`? I thought we replaced this with the slate decoration API?

msenejoa: as discussed, highlighter is a library plugin, unless we want to mount different instances of slatejs for each result

PR #274, opened 2020-06-24
Michael/bad ref

msenejoa:
* creates a new block if `BadRefId` error is thrown
* validates email for `Login`


packages/databyss-ui/components/Form/TextInputField.js
```diff
@@ -6,9 +6,16 @@ import {
 import { View, Text, TextInput } from '../../primitives'
 
 const TextInputField = React.forwardRef(
-  ({ path, label, placeholder, ...others }, ref) => {
+  ({ path, label, placeholder, hasError, ...others }, ref) => {
```
pzhine, line 9: Let's change `hasError` to `errorMessage` and show the error message in `uiTextSmall` underneath the field

PR #275, opened 2020-06-25
Mila/dropdown menu

mila-sevilla:


pzhine:
- [ ] Remove/comment out "End source" and "End location" for now
- [ ] Remove hotkey tips from menu items
- [ ] Pressing ESC key should close menu
- [ ] Highlight/active colors on Google results should match dropdown

packages/databyss-ui/components/Menu/DropdownContainer.js
```diff
@@ -0,0 +1,62 @@
...
+  pointerEvents: open ? 'auto' : 'none',
+  ...position,
+})
+
+const DropdownContainer = forwardRef(
+  (
+    {
+      children,
+      position = { top: 0 },
+      widthVariant = 'dropdownMenuMd',
```
pzhine, line 23: Let's use the static `DropdownContainer.defaultProps` for prop defaults.

packages/databyss-ui/components/Menu/DropdownListItem.js
```diff
@@ -0,0 +1,54 @@
...
+            mr="small"
+            color="text.2"
+          >
+            {menuItem.textSymbol}
+          </Text>
+        )}
+        {menuItem.icon && (
+          <Icon
+            sizeVariant="small"
+            width={pxUnits(20)}
```
pzhine, line 36: isn't the `width` specified by the size variant? maybe we need a new variant, like `menu`?

packages/databyss-ui/components/PageContent/ArchiveBin.js
```diff
@@ -30,16 +35,45 @@ export const ArchiveBin = ({ pages }) => {
...
+        <Icon sizeVariant="medium" color="text.1">
+          <MenuSvg />
+        </Icon>
+      </BaseControl>
+      {showMenu && (
+        <ClickAwayListener onClickAway={() => setShowMenu(false)}>
+          <DropdownContainer
+            open={showMenu}
+            position={{
+              top: 36,
```
pzhine, line 63: This won't scale relative to the size of the Icon/BaseControl. Can we make a constant or theme value for `menuLauncherSize`? I think this should be the same as `editorMarginMenuItemHeight`, so maybe we can just rename that one to be more generic...

packages/databyss-ui/theming/views.js
```diff
@@ -56,6 +56,15 @@ const widthVariants = {
   form: {
     maxWidth: pxUnits(550),
   },
+  dropdownMenuSm: {
```
pzhine, line 59: Despite the abbreviations in styled-system, let's use full names ("Medium", "Large", etc)...

packages/databyss-ui/components/Menu/DropdownListItem.js
mila-sevilla: Yeah, that was unnecessary to add since 20px is the size variant for small. I deleted it now :)

PR #279, opened 2020-06-25
new page logic

msenejoa:
* new page waits for server response before mounting
* `use-context-selector` in `PageProvider`

### DIFF
--- a/packages/databyss-api/src/models/Page.js
+++ b/packages/databyss-api/src/models/Page.js
@@ -1,8 +1,8 @@
 import mongoose from 'mongoose'
+import ObjectId from 'bson-objectid'
 import Entry from './Entry'
 import Block from './Block'
 import Selection from './Selection'
-import ObjectId from 'bson-objectid'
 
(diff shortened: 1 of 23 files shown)


packages/databyss-api/src/models/Page.js
```diff
@@ -38,18 +39,26 @@ const PageSchema = new Schema({
 /* eslint-disable prefer-arrow-callback */
 /* eslint-disable func-names */
 PageSchema.method('addEntry', async function(values = {}) {
+  const _entryId = ObjectId().toHexString()
```
pzhine, line 42: Do we need to generate the IDs explicitly? I think calling `Entry.create` and `Block.create` will generate them...

msenejoa: I can create the `id's` explicitly, or i can create and save the `entry`, create and save the `block`, then I would have to save the `entry` again with the new generated `block._id` since they're both linked.

PR #299, opened 2020-07-01
selection fix

### LINKED ISSUE #263: RichTextControl requires 2 clicks
**Observed behavior**
1. Create a source
3. Open edit source modal
4. Click into a rich text field (name or citation) that is not currently highlighted.
5. You have to click again to focus the cursor in that field

**Expected behavior**
Only one click should be necessary to focus the cursor in a RichTextControl

**Screenshots**
bug

### LINKED ISSUE #280: Delete Page should be inactive if Internet is Down
I believe this is a similar logic to New Page. Delete Page should only happen if there's a connection to the server.

msenejoa:
* fix selection bug in refactor
* fix entry search in refactor
* disable dismiss on source modal with empty field
* return search results if they have punctuation around the query
* fixes #280
* fixes #263


packages/databyss-ui/primitives/Control/native/RichTextInput.js
```diff
@@ -1,11 +1,34 @@
...
+    ref
+  ) => {
+    const editorRef = useRef(null)
+
     const _onBlur = event => {
       onBlur(event)
-      //  setTimeout(() => editor.blur(), 50)
+    }
+
+    useEffect(
```
pzhine, line 17: We should keep slate-specific code in `SingleLine`. How about we pass `active` into `SingleLine` instead of `setEditor` and move this `useEffect` inside?

packages/databyss-ui/primitives/Modal/ModalView.js
```diff
@@ -10,6 +10,7 @@ const ModalView = ({
   secondaryChild,
   onDismiss,
   children,
+  disabled,
```
pzhine, line 13: This naming is confusing, because `disabled` on the modal would imply that you can't do/change anything in the modal. Maybe rename this prop to `canDismiss`? Keep in mind we are currently allowing a click outside of the modal to dismiss as well, so you would have to prevent that if `canDismiss` is false.

PR #304, opened 2020-07-03
Paul/author headers

pzhine:


pzhine:
incorporated into #269

PR #312, opened 2020-07-08
Michael/clipboard refactor

msenejoa:


packages/databyss-editor/state/reducer.ts
```diff
@@ -74,13 +76,46 @@ export default (
...
+              ? state.selection.anchor.index
+              : state.selection.anchor.index + 1
+
+            // insert block at index
+            draft.blocks.splice(
+              _spliceIndex,
+              _isCurrentBlockEmpty ? 1 : 0,
+              ..._frag
+            )
+            draft.resetState = true
```
pzhine, line 105: What if we used our existing `operations` queue for this? So something like:
```
draft.operations.reloadAll = true
```

packages/databyss-editor/lib/clipboardUtils.js
```diff
@@ -0,0 +1,124 @@
+import { isAtomicInlineType } from './util'
```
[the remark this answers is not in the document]
msenejoa: yes, i will migrate this file over to typescript

packages/databyss-editor/lib/clipboardUtils.js
```diff
@@ -0,0 +1,124 @@
...
+  if (isSelectionCollapsed(state.selection)) {
+    return []
+  }
+
+  let frag = []
+
+  const { blocks, selection } = state
+  const { anchor, focus } = selection
+
+  let _blocks = cloneDeep(blocks)
```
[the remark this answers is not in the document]
msenejoa: a splice function will throw an error because `blocks` is immutable

packages/databyss-editor/state/EditorProvider.tsx
```diff
@@ -132,10 +137,41 @@ const EditorProvider: React.FunctionComponent<PropsType> = ({
...
+    // TODO: set plain text
+    e.clipboardData.setData('text/plain', 'Hello, world!')
+
+    // set application data for clipboard
+    e.clipboardData.setData('application/databyss-frag', JSON.stringify(_frag))
+
+    // TODO: SET HTML
+    dispatch({
+      type: COPY,
+      payload: {},
```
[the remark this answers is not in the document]
msenejoa: if `payload` is removed the reducer will throw an error at `let nextSelection = payload.selection` even if the reducer doesn't catch the `COPY` action in the switch statement.

packages/databyss-editor/state/reducer.ts
msenejoa: `draft.operations` is an array.

packages/databyss-editor/lib/slateUtils.js
```diff
@@ -144,3 +144,37 @@ export const toggleMark = (editor, format) => {
...
+    }
+    if (node.italic) {
+      _children = `<i>${_children}</i>`
+    }
+    return _children
+  }
+
+  const children = node.children.map(n => serialize(n)).join('')
+
+  switch (node.type) {
```
pzhine, line 164: We will need to create a mockup and spec for the HTML export of sources and topics. We may even want to link them to routes in the app. Let's discuss next week.

packages/databyss-editor/state/EditorProvider.tsx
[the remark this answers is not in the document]
pzhine: Ah ok, we should probably change that to `let nextSelection = payload?.selection` in the reducer to avoid this. According to the [FSA spec](https://github.com/redux-utilities/flux-standard-action#actions), payload should be optional.

packages/databyss-editor/state/EditorProvider.tsx
```diff
@@ -132,10 +147,64 @@ const EditorProvider: React.FunctionComponent<PropsType> = ({
...
+        payload: {
+          data,
+        },
+      })
+      return
+    }
+
+    if (plainTextDataTransfer) {
+      const data = plainTextToDatabyssFrag(plainTextDataTransfer)
+      dispatch({
```
pzhine, line 191: This looks the same as above, can we consolidate?

packages/databyss-editor/state/reducer.ts
```diff
@@ -78,9 +86,103 @@ export default (
...
+        case PASTE: {
+          const _frag = payload.data
+
+          // check if paste is occuring on an atomic block
+
+          const {anchor: _startAnchor} = sortSelection(state.selection)
+          const _startBlock = state.blocks[_startAnchor.index]
+
+          // if selection in the middle of atomic prevent paste
+          if(!isSelectionCollapsed(state.selection) || _startBlock.text.textValue.length !== _startAnchor.offset){
```
pzhine, line 105: What about if it's at the beginning of the atomic? In that case, I think we should paste the fragment before the atomic...
In fact, in these cases where it's in the middle or end, we should paste the fragment after the atomic instead of preventing it.

packages/databyss-editor/state/reducer.ts
```diff
@@ -78,9 +86,103 @@ export default (
...
+                anchor: { index: _selectionIndex, offset: _offset },
+                focus: { index: _selectionIndex, offset: _offset },
+              }
+              nextSelection = _nextSelection
+            } else {
+              // merge fragment at current block
+              const { blocks, selection } = draft
+              const { anchor } = selection
+              const _index = anchor.index
+              const _mergedBlock = insertBlockAtIndex({
```
pzhine, line 158: Ah, ok, this function's purpose is clearer here. So a better signature might be 
```
insertText(
  { block, text, offset } : 
  { block: Block, text: Text, offset: number }
)
``` 
and you'd call it like
```
insertText({ 
  blocks: blocks[anchor.index], 
  text: frag[0].text, 
  offset: selection.anchor.offset,
})
```

packages/databyss-editor/state/reducer.ts
```diff
@@ -78,9 +86,103 @@ export default (
...
+                block: blocks[_index],
+                blockToInsert: _frag[0],
+                index: anchor.offset,
+              })
+
+              // insert new block at current index
```
pzhine, line 164: This should be "replace block at current index with merged block", right?

packages/databyss-editor/state/reducer.ts
```diff
@@ -78,9 +86,103 @@ export default (
...
+            if( isAtomicInlineType(_startBlock.type)){
+              break
+            }
+          }
+
+          if (!isSelectionCollapsed(state.selection)) {
+            deleteBlocksAtSelection({ state: state, draftState: draft })
+          }
+
+          if (_frag.length) {
```
pzhine, line 115: Will frag.length ever be 0? if so, let' replace this with a guard close to bail if frag is empty

packages/databyss-editor/lib/clipboardUtils.tsx
```diff
@@ -0,0 +1,445 @@
...
+  }
+
+  return {
+    anchor: _anchor,
+    focus: _focus,
+  }
+}
+
+// returns fragment in state selection
+export const getCurrentSelection = (state: EditorState): Block[] => {
```
pzhine, line 134: let's call this `getFragmentForCurrentSelection`

packages/databyss-editor/lib/clipboardUtils.tsx
```diff
@@ -0,0 +1,445 @@
...
+    const _selectionLength = focus.offset - anchor.offset
+    const _block = blocks[anchor.index]
+    // split block at anchor offset and use `after`
+    const _firstSplit = splitBlockAtOffset({
+      block: _block,
+      offset: anchor.offset,
+    }).after
+
+    // split block at length of selection and get `before`
+    const _secondSplit = splitBlockAtOffset({
```
pzhine, line 158: Why do we need 2 splits here?

packages/databyss-editor/lib/clipboardUtils.tsx
```diff
@@ -0,0 +1,445 @@
...
+      ? { after: blocks[anchor.index] }
+      : splitBlockAtOffset({
+          block: blocks[anchor.index],
+          offset: anchor.offset,
+        })
+
+    if (firstBlock) {
+      frag.push({
+        ...firstBlock,
+        _id: getId(firstBlock.type, blocks[anchor.index]._id),
```
pzhine, line 184: Maybe a better name for this fn is `fragmentId()`

packages/databyss-editor/lib/clipboardUtils.tsx
```diff
@@ -0,0 +1,445 @@
...
+    const _sliceLength = focus.index - anchor.index
+
+    if (_sliceLength > 1) {
+      _blocks.splice(anchor.index + 1, _sliceLength - 1).forEach((b: Block) => {
+        frag.push({ text: b.text, type: b.type, _id: getId(b.type, b._id) })
+      })
+    }
+
+    // get in between frags
+    const { before: lastBlock } = splitBlockAtOffset({
```
pzhine, line 197: What if the last block is atomic?

packages/databyss-editor/lib/clipboardUtils.tsx
```diff
@@ -0,0 +1,445 @@
...
+    if (firstBlock) {
+      frag.push({
+        ...firstBlock,
+        _id: getId(firstBlock.type, blocks[anchor.index]._id),
+      })
+    }
+
+    const _sliceLength = focus.index - anchor.index
+
+    if (_sliceLength > 1) {
```
pzhine, line 190: This is confusing. To handle all the blocks covered by the selection, can't we just use a forEach loop like:
```
blocks.forEach((block, index) => {
  if (index === anchor.index) {
    // do firstBlock stuff
  } else if (index === focus.index) {
    // do lastBlock stuff
  } else if (index > anchor.index && index < focus.index) {
    // do in between block stuff
  }
}
```

packages/databyss-editor/lib/clipboardUtils.tsx
```diff
@@ -0,0 +1,445 @@
...
+      })
+    }
+  }
+  // add metadata
+  frag = addBlockData(frag)
+
+  return frag
+}
+
+const mergeBlocks = ({
```
pzhine, line 215: Let's call this `mergeText` with the following signature: 
```
mergeText(a: Text, b: Text): Text
```

pzhine:
Please rebase to `master`

packages/databyss-editor/state/EditorProvider.tsx
```diff
@@ -132,10 +147,64 @@ const EditorProvider: React.FunctionComponent<PropsType> = ({
...
+      'application/x-databyss-frag'
+    )
+
+    const plainTextDataTransfer = e.clipboardData.getData('text/plain')
+
+    // TODO: html parser for rich text
+
+    if (databyssDataTransfer) {
+      let data = JSON.parse(databyssDataTransfer)
+      data = resetIds(data)
```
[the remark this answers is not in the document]
msenejoa: If we dont reset the `ids` on every paste, we will end up duplicating block `ids` when pasting multiple times with the same payload

packages/databyss-editor/lib/clipboardUtils.tsx
```diff
@@ -0,0 +1,445 @@
+import ReactDOMServer from 'react-dom/server'
+import ObjectId from 'bson-objectid'
+import cloneDeep from 'clone-deep'
+import { BlockType } from '@databyss-org/services/interfaces'
+import { Text, Range, Selection, EditorState, Block } from '../interfaces'
+import { isAtomicInlineType } from './util'
+import { stateToHTMLString } from './slateUtils'
+
+interface BasicBlock {
```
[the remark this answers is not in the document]
msenejoa: when doing block manipulation (split, merge) we do not require the block `id`, the block `id` logic is handled by the functions calling `splitBlockAtOffset` `mergeBlock` `insertText`

packages/databyss-editor/lib/clipboardUtils.tsx
```diff
@@ -0,0 +1,445 @@
...
+    return []
+  }
+
+  let frag: Block[] = []
+
+  const { blocks, selection } = state
+
+  const { anchor, focus } = sortSelection(selection)
+
+  const _blocks = cloneDeep(blocks)
```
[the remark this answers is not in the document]
msenejoa: `blocks` is immutable and we do a `blocks.splice` later in the function to get the block fragment

packages/databyss-editor/lib/clipboardUtils.tsx
msenejoa: "this is an `|(first split)`example of a selected `|(second split)`fragment"

the selected fragment is `example of a selected`

packages/databyss-editor/lib/clipboardUtils.tsx
msenejoa: `splitBlockAtOffset` handles this case

packages/databyss-editor/lib/clipboardUtils.tsx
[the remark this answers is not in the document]
pzhine: Ah ok. Let's make `_id` an optional field so we can use the `Block` interface here. As in
```
export interface Block {
  _id?: string
  type: BlockType
  text: Text
}
```

While we're at it, we can do the same for `Selection`, because TS is complaining sometimes that `_id` is missing.

packages/databyss-editor/lib/clipboardUtils.tsx
```diff
@@ -0,0 +1,476 @@
...
+      }),
+      type: 'ENTRY',
+    }
+  }
+  return mergedBlock
+}
+
+export const deleteBlocksAtSelection = ({
+  state,
+  draftState,
```
pzhine, line 276: Why do we need `draftState` and `state`. What if we just passed it the draft state and it read/write everything from there?

packages/databyss-editor/package.json
```diff
@@ -36,7 +36,7 @@
     "is-hotkey": "^0.1.5",
     "lodash": "^4.17.15",
     "lorem-ipsum": "^2.0.3",
-    "slate-react": "0.57.2",
+    "slate-react": "0.58.4",
```
msenejoa, line 39: this change causes the cursor to jump, an issue we fixed with our fork of `slate`

PR #324, opened 2020-07-15
spellcheck debounce

### LINKED ISSUE #292: Searching for Words with Accents
Right now you can't type accents into the search line, you should be able to. But more importantly, the search should find a word with an accent, even if that accent wasn't placed there. For example, if you type epoux, you should be able to find époux.

### LINKED ISSUE #303: Spellcheck Flashes
We need to make spell check only implement itself at the end of every word. Right now it flashes on every misspelled word in a block.

ALSO, turn spellcheck off for source blocks

### LINKED ISSUE #314: Apostrophes in Search Box
You should be able to type apostrophes in search box. Because you can't, you can't search for words with apostrophes right now.

msenejoa:
* fixes #303
* fixes #292
* fixes #314


pzhine:
Search isn't finding some entries now.

master
Screenshot 2020-07-20 01 16 15

this branch
Screenshot 2020-07-20 01 15 51

pzhine:
Please rebase to `master`

PR #334, opened 2020-07-25
Mila/redesign homepage

mila-sevilla:


packages/databyss-ui/modules/Homepage/Hero/Navbar.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react'
+import { BaseControl, View, Text } from '@databyss-org/ui/primitives'
+import theme, { timing } from '@databyss-org/ui/theming/theme'
+import { pxUnits } from '@databyss-org/ui/theming/views'
+
+const Navbar = ({ lightTheme, navLinks }) => {
+  // const location = useLocation()
```
pzhine, line 7: We don't need routing for now on the homepage app.

packages/databyss-ui/modules/Homepage/homepageContent.json
```diff
@@ -0,0 +1,101 @@
...
+            "title": "Databyss",
+            "headline": "You research, dabble, experiment, take notes, and get lost in your thoughts. Databyss is your new word-processor.",
+            "buttonText": {
+                "bold": "Sign up",
+                "normal": "(for free)"
+            },
+            "buttonHref": "https://app.databyss.org/signup",
+            "navLinks": [
+                {
+                    "name": "Home",
```
pzhine, line 23: We don't need the Home link here. However, we should like the icon on `Private.js` back to `www.databyss.org` so it goes to the homepage.

packages/databyss-ui/theming/mediaBreakpoints.js
```diff
@@ -0,0 +1,4 @@
+export const mobileBreakpoint = { maxWidth: 425 }
```
pzhine, line 1: We already have a `/theming/responsive` file that looks similar. Can we adjust and use that?

PR #335, opened 2020-07-28
header

msenejoa:
* editor sticky header with current source/topic


packages/databyss-ui/components/PageContent/PageContent.js
```diff
@@ -69,31 +63,34 @@ const PageContainer = React.memo(({ anchor, id, page }) => {
...
-        alignItems="center"
-        flexDirection="row"
-        justifyContent="space-between"
-      >
-        <PageHeader
-          ref={headerRef}
-          pageId={id}
-          onNavigateDownFromHeader={onNavigateDownToEditor}
+    <View height="100vh" overflow="scroll" ref={editorWindowRef}>
+      <PageSticky page={editorState} />
```
pzhine, line 72: I believe we discussed this, but let's make this prop just the list of "rendered" breadcrumbs in the path. This way, we can keep the logic that requires the editor state within the EditorProvider where it can access it directly, and just "publish" the rendered path elements on the ref. I think this is preferable to passing the whole state around.

msenejoa: I just moved the `pagePath` logic to `EditorProvider`. It now passes a `string` instead of the whole state. I'm not sure what you mean by `rendered`.

packages/databyss-editor/state/EditorProvider.tsx
```diff
@@ -13,6 +18,7 @@ import {
 import { Text, Selection, EditorState } from '../interfaces'
 import initialState from './initialState'
 import reducer from './reducer'
+import { getPagePath } from '../../databyss-ui/components/PageContent/_helpers'
```
pzhine, line 21: This method should be moved to a lib in the editor package

packages/databyss-ui/components/PageContent/_helpers.ts
```diff
@@ -0,0 +1,59 @@
...
+import { EditorState } from '../../../databyss-editor/interfaces/index'
+import { stateBlockToHtmlHeader } from '@databyss-org/editor/lib/slateUtils.js'
+
+const getBlockPrefix = (type: string): string =>
+  ({
+    SOURCE: '@',
+    TOPIC: '#',
+  }[type])
+
+export const getPagePath = (page: EditorState): string => {
```
pzhine, line 14: I think this should return an array of strings representing the paths. That way, the rendering component can determine the padding between path elements and do truncation or ellipsis as needed.

PR #362, opened 2020-08-05
mjanson/parse pdf annotations

jansensan:


packages/databyss-ui/components/PDFDropZone/PDFDropZoneManager.js
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    return file
+  }
+
+  const toDatabyssBlocks = (fileName, annotations) => {
+    // create response object with source as first item
+    const response = [
+      {
+        _id: new ObjectId().toHexString(),
```
pzhine, line 97: Now that I think about it, we re-generate all `_id`s on an `insert` operation for ENTRY blocks, so let's try leaving this out for clarity.

packages/databyss-ui/components/PDFDropZone/PDFDropZoneManager.js
```diff
@@ -0,0 +1,244 @@
...
+    event.stopPropagation()
+    event.preventDefault()
+
+    setParsed(false)
+
+    const file = getFileToProcess(event)
+
+    if (!file) {
+      setDropAreaVisibility(false)
+      showAlert('⚠️ Unable to parse document', 'Ensure to use a PDF document.')
```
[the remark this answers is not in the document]
jansensan: Then I would suggest this:

heading: "⚠️ Unable to import file" (as you suggested)
message "We are only able to import PDF files at this time. Please ensure to use a PDF document."

Good error messaging is about guiding users about what to do, not only about what has happened. Simply saying "we can't" leaves users in a dead end. With the suggestions, users know what they can do. Let's be clear, not leave users to read in between the lines.

packages/databyss-ui/components/PDFDropZone/PDFDropZoneManager.js
jansensan: So I simply remove that  prop?

PR #375, opened 2020-08-14
Mila/page content width and fix overflow issues

### LINKED ISSUE #323: Apply widthVariant="content" to editor page
Right now, the text reaches to the extent of the browser, which makes it harder to read on wider screens:
Screenshot 2020-07-14 23 57 31

Adding `widthVariant="content"` should fix this.

### LINKED ISSUE #347: Long page title creates horizontal scrollbar in pages sidebar
**Observed behavior**
1. Give a page a long enough title that it will wrap in the pages sidebar
2. Pages sidebar shows scrollbar as shown. Note that if you don't have a mouse plugged in, the scrollbar will be hidden until you scroll right or left.
3. Also note that the scrollable area doesn't extend to the bottom of the sidebar.

Screenshot 2020-08-02 14 22 02

**Expected behavior**
Page title should wrap well before the right extent of the sidebar, meaning a horizonal scroll should never be possible and there should be some padding between the rightmost character of the title and the extent of the sidebar.

Also, extend scrollable area vertically so it fills the container.

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

### LINKED ISSUE #348: Long entry in multiline rich text control can break horizontal boundaries
**Observed behavior**
1. Open a source modal
2. Paste a really long, unbroken string into the "citation" field
3. The string extends offscreen to the right of the field, creating a scrollbar in the modal

Screenshot 2020-08-01 22 08 22

**Expected behavior**
Contents of a multiline text control should not overflow the right boundary. Probably can be solved with an `overflow: hidden` or maybe `overflow: scroll`

mila-sevilla:
Fix #323: Change max-width of content to 700px and fix overflow problem of long text
Fix #348: Fix also overflow on text input in modal
Fix #347: Fix issues with long page titles in sidebar

### DIFF
--- a/packages/databyss-ui/modules/Modals/TopicModal.js
+++ b/packages/databyss-ui/modules/Modals/TopicModal.js
@@ -7,12 +7,10 @@ import ValueListProvider, {
 import {
   ModalWindow,
   View,
-  Grid,
   TextControl,
   List,
 } from '@databyss-org/ui/primitives'
 import { useNavigationContext } from '@databyss-org/ui/components/Navigation/NavigationProvider/NavigationProvider'
-import { widthVariants, pxUnits } from '@databyss-org/ui/theming/views'
 
 const ControlList = ({ children, ...others }) => (
   <List horizontalItemPadding="small" {...others}>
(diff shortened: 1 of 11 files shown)


packages/databyss-ui/modules/Modals/TopicModal.js
```diff
@@ -61,7 +62,9 @@ const TopicModal = ({ refId, visible, onUpdate, id }) => {
               <Grid>
                 <View
                   paddingVariant="none"
-                  widthVariant="content"
+                  maxWidth={`calc(${widthVariants.form.maxWidth} - ${pxUnits(
```
pzhine, line 65: This seems awkward, there must be a better way we can do this....

PR #376, opened 2020-08-14
Fetched additional metadata from Crossref

jansensan:
- Fetched additional metadata from Crossref
- Added modal for user to edit metadata before inserting
- Converted entry block type to source
- Post-rebase corrections

### DIFF
--- a/packages/databyss-services/interfaces/Block.ts
+++ b/packages/databyss-services/interfaces/Block.ts
@@ -28,7 +28,10 @@ export interface Citations {
 
 export interface SourceDetail {
   authors: Author[]
-  citations: Citations[]
+  citations?: Citations[]
+  doi?: string
+  issn?: string
+  year?: number
 }
 
 export interface SourceCitationHeader extends Source {}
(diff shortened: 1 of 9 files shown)


packages/databyss-services/interfaces/Block.ts
```diff
@@ -28,7 +28,10 @@ export interface Citations {
 
 export interface SourceDetail {
   authors: Author[]
-  citations: Citations[]
+  citations?: Citations[]
```
pzhine, line 31: Shouldn't this be `Citation[]`?

packages/databyss-ui/components/PDFDropZone/PDFDropZoneManager.js
```diff
@@ -35,9 +36,71 @@ const viewStyles = () => ({
 const StyledView = styled(View, viewStyles)
 
 // methods
+const buildSource = data => {
```
pzhine, line 39: Arg `data` and return value should have the shape defined by the `Source` interface: https://github.com/databyss-org/databyss/blob/master/packages/databyss-services/interfaces/Block.ts#L36

packages/databyss-ui/modules/Modals/MetadataModal.js
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  // render methods
+  const renderAuthorFields = () => {
+    const labelStr = values.authors.length > 1 ? 'Authors' : 'Author'
+    return (
+      <View>
+        <Legend variant="uiTextNormalSemibold">{labelStr}</Legend>
+        {values.authors.map((author, index) => (
+          <div key={index}>
```
pzhine, line 112: I think we want a `<Grid singlerow>` here, for side-by-side children.

PR #387, opened 2020-08-19
Mila/new combined search

### LINKED ISSUE #400: Source and Topic index page formatting
- [x] Apply standard body content maxWidth to the content
- [x] Fix block alignment to match search results

**Screenshots**
Screen Shot 2020-08-22 at 9 14 37 AM

mila-sevilla:
- Combining the search, so you can search for entries, authors, topics, sources and page titles together, instead of separately
- Fix correct highlighting of active sidebar item 
- Fix decoding of special characters in url
- Fix #400 
- Refactor MakeLoader component to also accept an array of resources


packages/databyss-ui/components/PageContent/IndexPageContent.js
```diff
@@ -1,9 +1,18 @@
...
-  <ScrollView p="medium" flex="1" maxHeight="98vh">
+  <ScrollView
+    p="medium"
+    flex="1"
+    maxHeight="98vh"
+    maxWidth={widthVariants.content.maxWidth}
+    css={{
+      overflowWrap: 'anywhere',
+      wordBreak: 'break-word',
+    }}
```
pzhine, line 14: Let's add these styles to a `wrapVariant` on all `View` and derived components

packages/databyss-ui/components/PageContent/IndexPageContent.js
```diff
@@ -1,9 +1,18 @@
...
-import { pxUnits } from '@databyss-org/ui/theming/views'
+import { pxUnits, widthVariants } from '@databyss-org/ui/theming/views'
 
 const IndexPageContent = ({ title, children }) => (
-  <ScrollView p="medium" flex="1" maxHeight="98vh">
+  <ScrollView
+    p="medium"
+    flex="1"
+    maxHeight="98vh"
+    maxWidth={widthVariants.content.maxWidth}
```
pzhine, line 10: why not just use `widthVariant="content"`?

packages/databyss-ui/components/SearchContent/SearchResults.js
```diff
@@ -0,0 +1,46 @@
...
+  Text,
+  View,
+  BaseControl,
+  Grid,
+  Icon,
+} from '@databyss-org/ui/primitives'
+import { widthVariants } from '@databyss-org/ui/theming/views'
+
+export const SearchResultsContainer = ({ children }) => (
+  <View mb="medium" maxWidth={widthVariants.content.maxWidth}>
```
pzhine, line 12: why not `widthVariant="content"`?

packages/databyss-ui/modules/Sidebar/Sidebar.js
```diff
@@ -30,14 +30,19 @@ Section.defaultProps = {
   variant: 'heading3',
 }
 
+const padding = 26
+const headerHeight = 34
+const footerHeight = 48
+const searchBar = 54
+
+export const sidebarListHeight = `calc(100vh - ${pxUnits(
```
pzhine, line 38: I see why we're doing this, but we should make this more responsive. I think we should use flexbox to expand the navbar and sidebar vertically rather than positioning the new page buttons absolutely. Then we can let the flex layout do the work of keeping everything flush and aligned. We can do this in a separate PR if you like.

packages/databyss-ui/primitives/Control/native/TextInput.js
```diff
@@ -31,6 +32,7 @@ const TextInput = forwardRef(
         onChange={e => onChange({ textValue: e.target.value })}
         ref={ref}
         css={[resetCss].concat(concatCss)}
+        autoFocus={autoFocus}
```
pzhine, line 35: What does this prop do?

packages/databyss-ui/components/Navigation/NavigationProvider/NavigationProvider.js
```diff
@@ -65,7 +64,7 @@ const NavigationProvider = ({ children }) => {
     return getTokensFromPath().type ? getTokensFromPath().type : 'pages'
   }
 
-  const getQueryParams = () => queryString.parse(location.search)
+  const getQueryParams = () => decodeURIComponent(location.search)
```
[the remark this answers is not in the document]
mila-sevilla: That didn't work unfortunately, that's why I changed it. I don't know why but it didn't decode the string properly, it just left all the special characters

packages/databyss-ui/components/PageContent/IndexPageContent.js
mila-sevilla: it didn't work, it just translated the css to "max-width: content" and not "max-width: 700px"

packages/databyss-ui/components/SearchContent/SearchResults.js
mila-sevilla: Same reason as with the other component, it doesn't get translated to pixels anymore

packages/databyss-ui/primitives/Control/native/TextInput.js
mila-sevilla: It autofocuses an input: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefautofocus. I used it for when we switch route to /search so it autofocus the search input and the user can start typing right away

packages/databyss-ui/modules/Sidebar/Sidebar.js
mila-sevilla: Yeah I tried a couple of times to refactor this part but it's pretty difficult because the new page button is part of <SidebarCollpased> but it also needs to cover the <Sidebar> component, so I don't know if that's possible without position absolute. I think it probably makes sense to try in another PR...

packages/databyss-ui/components/PageContent/IndexPageContent.js
mila-sevilla: Actually I tried it again and strangely it works now... maybe I passed the `maxWidth` prop before instead of `widthVariant`, I don't know

packages/databyss-ui/components/PageContent/IndexPageContent.js
mila-sevilla: That makes sense, should the overflowWrap: 'anywhere' be the default then?

PR #395, opened 2020-08-20
bug fix

### LINKED ISSUE #342: Spellcheck When Moving to New Block
Logic is wrong when you're toggling.

### LINKED ISSUE #346: 2 Clipboard Fixes
1) You should be able to paste if you've highlighted text with an atomic in it.

2) All empty entries should be removed on a paste from an external app (they should remain if you copy/paste internally)

msenejoa:
* fixes #346 
* fixes #342

### DIFF
--- a/packages/databyss-ui/components/SourcesContent/IndexSourceContent.js
+++ b/packages/databyss-ui/components/SourcesContent/IndexSourceContent.js
@@ -24,40 +24,44 @@ const IndexSourceContent = ({ relations }) => {
     navigate(`/pages/${pageId}#${entryId}`)
   }
 
-  const _results = Object.keys(relations.results).map((r, i) => (
-    <View key={i} mb="medium">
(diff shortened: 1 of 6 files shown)


packages/databyss-ui/components/SourcesContent/IndexSourceContent.js
```diff
@@ -24,40 +24,44 @@ const IndexSourceContent = ({ relations }) => {
...
-              <PageSvg />
-            </Icon>
-            <Text variant="bodyHeading3">{pages[r].name}</Text>
-          </Grid>
-        </BaseControl>
-      </View>
+  const _results = Object.keys(relations.results)
+    // filter out results not associated to a page
+    // page may have been archived
+    .filter(r => pages[r]?.name)
```
msenejoa, line 30: filter out results not associated with a page

PR #417, opened 2020-09-01
Mila/fix tooltip issue and only show on mouseup

### LINKED ISSUE #407: Tooltip Error has Returned
When you highlight right to left and then click a markup (loc, bold, italics), the tooltip moves to the other position.

We fixed this on master, but I believe that Paul's highlight/scroll fix regressed back to this error.

mila-sevilla:
fix #407 

Show formatting tooltip only on mouseup and fix the jumping around after you try to format a text that you selected backwards.


pzhine:
Hey @mila-sevilla, I noticed you made some changes to the sidebar code as well in this PR. It seems to have caused a regression in the layout when the sidebar is collapsed.

Screen Shot 2020-09-10 at 1 24 30 PM

pzhine (changes requested):
See sidebar regression in PR comments.

PR #429, opened 2020-09-08
Adapted readonly view to mobile

jansensan:
Adapted readonly view to mobile
- Ensured PDFDropZoneManager is inactive when public
- Ensured sidebar and panels behave and are displayed properly


pzhine:
Public mobile will use the same app as private mobile, so closing

PR #432, opened 2020-09-08
fix

msenejoa:
* allows returns in the middle of an entry
* Trim empty line after double enter in the middle of entry with multiple lines: it creates a new block, and cursor is in front of the first word of that new block
* `#` and `@` should not function if a block has multiple lines
* If you do a double enter right before the first letter of a line, it should now create a new block


pzhine:
superseded by #437

PR #434, opened 2020-09-09
Michael/optimize pass

msenejoa:
* add block metadata to blocks before they reach reducer
* improve performance page to include ranges
* move functions inside of `useMemo` on `ContentEditable`


pzhine:
superseded by #436

PR #497, opened 2020-09-21
Michael/bugfix 15

### LINKED ISSUE #484: Force email address to lowercase before sending to db
We want to avoid a situation where a user has two accounts because of capitalization differences in their email, as in
`john.smith@gmail.com` and `John.Smith@gmail.com`

### LINKED ISSUE #490: Doubling of Search Term
If you search a term, then click on an entry under that term, and then click back to the search results, the term you searched for will be doubled. This seems to be because of the slate highlighting.

### LINKED ISSUE #493: Read-Only Margins on Desktop Mess Up
On Read-Only URLs that are made public, or the Read-Only Archive pages, the margins are messed up. Definitely too far to the left. 

It should look exactly like the regular pages.

### LINKED ISSUE #494: Archiving Your Default Page
If you Archive your default page, and then delete forever another page--the default page will still come up, but as read-only.

I haven't even checked what happens when you delete forever the default--probably something better.

### LINKED ISSUE #495: Fixing Archive Structure
1) You can still search archived pages. Archived pages shouldn't show up in the search.

2) Topics and sources that have been archived or deleted forever (and are the last versions of themselves) should not show up in the suggestions box.

### LINKED ISSUE #498: Typing in Search Doesn't Always Launch Search
If you tab into the search bar, and you are on a pages, sources, or topic sidebar view, anything you type will not be searched.

msenejoa:
* fixes #495
* fixes #494 
* fixes #490 
* fixes #484
* fixes #493
* fixes #498


packages/databyss-editor/components/Suggest/SuggestTopics.js
```diff
@@ -39,8 +39,10 @@ const SuggestTopics = ({ query, dismiss, onSuggestionsChanged }) => {
   }
 
   const onTopicsLoaded = topicsDict => {
-    const _topics = Object.values(topicsDict)
-    setSuggestions(_topics)
+    if (!suggestions) {
```
msenejoa, line 42: this would cause a `useEffect` browser error

PR #542, opened 2020-09-29
Michael/edit source modal

msenejoa:
# suggested sources tests 
1) If there are no suggestions the suggestion menu works
2) You can type multiple words in any order to search suggestions
3) Suggestions don't break with strange characters (i.e. "[" "\" etc.)
4) If you click on a suggestion this doesn't create a new topic


jansensan:
@msenejoa Closing this PR, your work has been included in https://github.com/databyss-org/databyss/pull/502

PR #560, opened 2020-10-08
Parsed additional data from catalog sources

jansensan:


jansensan:
Closing, much more work has been done around citations in #568

PR #568, opened 2020-10-20
Mjanson/edit source form citation service

jansensan:
This could replace #560, as it expands on that PR and adds the citation service.


jrceno:
I noticed 3 things. 1) The one error that I saw is still there (with a certain type of old source). 2) If you just type a source without pulling it from somewhere, opening the modal breaks immediately. 3) Why is the year being placed in the field year but not being placed in the citation?

PR #640, opened 2020-12-14
Michael/use pouch

pzhine:


packages/databyss-services/session/clientStorage.js
```diff
@@ -26,3 +26,15 @@ export function getAccountId() {
 export function deleteAccountId() {
   localStorage.removeItem('account')
 }
+
+export function setDefaultPageId(value) {
```
pzhine, line 30: Why are we storing this locally now?

packages/databyss-services/database/_helpers.ts
```diff
@@ -0,0 +1,90 @@
...
+import { PatchBatch } from '../interfaces/Patch'
+import { runPatches } from './patches'
+import { DocumentType, DbPage, MangoResponse } from './interfaces'
+import { PageHeader } from '../interfaces/Page'
+import { initSelection, initBlock, initPage } from './initialState'
+
+export const populatePage = async (_id: string): Promise<Page | null> => {
+  const _response = await db.find({
+    selector: {
+      documentType: 'PAGE',
```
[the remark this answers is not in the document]
msenejoa: `_type` is not allowed on PouchDB as a property

packages/databyss-data/schemas/blockRelationSchema.ts
```diff
@@ -0,0 +1,71 @@
...
+    },
+    modifiedAt: {
+      type: 'number',
+    },
+    createdAt: {
+      type: 'number',
+    },
+  },
+  required: [
+    '_id',
```
pzhine, line 59: Similarly, we should have a mixin for the `required` array so we can do
```
required: [
  ...pouchDocRequired,
  block,
  ...
]
```

packages/databyss-data/schemas/entrySchema.ts
```diff
@@ -0,0 +1,53 @@
...
+          items: {
+            type: 'string',
+          },
+        },
+      },
+    },
+    _id: {
+      type: 'string',
+    },
+    type: {
```
pzhine, line 27: Following the mixin pattern, we should have a `blockSchema` mixin for schemas that share the `block` base class.
```
export const blockSchema = {
  type: {
    type: 'string',
  },
  text: {
    $ref: 'text',
  },
  account: {
    type: 'string',
  },
}
```

packages/databyss-data/schemas/test.json
```diff
@@ -0,0 +1,36 @@
+{
```
pzhine, line 1: Can this be cleaned up or moved?

packages/databyss-editor/lib/clipboardUtils/deleteBlocksAtSelection.ts
```diff
@@ -124,6 +124,7 @@ export default (draft: EditorState) => {
   // collapse selection
   const _cursor = { offset: anchor.offset, index: anchor.index }
   draft.selection = {
+    $type: DocumentType.Selection,
```
pzhine, line 127: If we have to use `$type` or `DocumentType` outside of `@databyss-org/data` we've done something wrong. It is meant as a field internal to our database library for managing types. Let's discuss why we are using this here and some alternatives.

packages/databyss-services/interfaces/Block.ts
```diff
@@ -2,6 +2,13 @@ import { BlockRelation } from '@databyss-org/editor/interfaces/index'
 import { SelectOption } from './UI'
 import { Text } from './Text'
 
+export enum DocumentType {
```
pzhine, line 5: See note above about `DocumentType` scope. This should be in `@databyss-org/data`

packages/databyss-ui/components/Notify/NotifyProvider.js
```diff
@@ -47,7 +43,7 @@ class NotifyProvider extends React.Component {
   constructor(props) {
     super(props)
     startBugsnag(props.options)
-    this.shouldCheckOnlineStatus = props.shouldCheckOnlineStatus
+    // this.shouldCheckOnlineStatus = props.shouldCheckOnlineStatus
```
pzhine, line 46: TODO: restore online status checking and inspect pouchdb sync state so we can inform the user of sync progress

packages/databyss-ui/components/PageContent/PageHeader.js
```diff
@@ -34,13 +34,14 @@ const PageHeader = forwardRef(({ pageId, onNavigateDownFromHeader }, ref) => {
...
   const throttledAutosave = useCallback(
-    throttle((val) => {
+    debounce((val) => {
       const _pageData = {
         name: val.textValue ? val.textValue : noPageTitle,
         _id: pageId,
       }
       setPageHeader(_pageData)
-    }, process.env.SAVE_PAGE_THROTTLE),
+      // TODO: CHANGE THIS BACK TO process.env.SAVE_PAGE_THROTTLE
```
pzhine, line 43: Can we do this now?

PR #676, opened 2021-02-22
inline atomic fix

msenejoa:


packages/databyss-services/session/clientStorage.js
```diff
@@ -45,11 +50,27 @@ export const getAccountId = async () => {
...
+
+  // replace default page id on correct group for user
+  const _result = await findOne({
     $type: DocumentType.UserPreferences,
-    _id: 'user_preferences',
-    doc: { defaultPageId: value },
+    query: { _id: 'user_preference' },
   })
+
+  if (_accountFromLocation && _result) {
```
msenejoa, line 61: default page id is now a property in the groups array

PR #739, opened 2021-04-20
set inline topic bug fix

msenejoa:


packages/databyss-editor/components/ContentEditable.js
```diff
@@ -132,6 +133,9 @@ const ContentEditable = ({
   // if new atomic block has been added, save atomic
   useEffect(() => {
     if (state.newEntities.length) {
+      // flush the queue processor in order to get up to date values
+      EM?.process()
```
msenejoa, line 137: flush the queue

PR #742, opened 2021-04-22
Michael/inline atomics

msenejoa:


packages/databyss-data/pouchdb/sources/lib/setSource.ts
```diff
@@ -35,6 +41,48 @@ export const setSource = async (data: Source) => {
     return
   }
 
+  for (const _pageId of _relation!.pages) {
```
pzhine, line 44: this looks nearly identical to `setTopic`, so let's merge them

packages/databyss-editor/components/Suggest/SuggestSources.js
```diff
@@ -103,6 +153,56 @@ const SuggestSources = ({
 
   const _mode = resultsMode || LOCAL_SOURCES
 
+  // TODO: this function is generic for topic and suggestion
```
pzhine, line 156: let's make this generic now

packages/databyss-editor/lib/clipboardUtils/adjustSelectionToIncludeInlineAtomics.ts
```diff
@@ -25,15 +26,25 @@ export default ({
   }).filter(
     (r) =>
       r.marks.length &&
-      r.marks.filter((i) => Array.isArray(i) && i[0] === 'inlineTopic').length
+      r.marks.filter(
```
pzhine, line 29: you need the `.length` here, because `filter` still returns truthy if no results are found, eg:
`!![].filter(k => false) // === true`

packages/databyss-editor/lib/clipboardUtils/adjustSelectionToIncludeInlineAtomics.ts
```diff
@@ -25,15 +26,25 @@ export default ({
...
+      )
   )
   const _inlineRangesAtFocus = getRangesAtPoint({
     blocks,
     point: focus,
   }).filter(
     (r) =>
       r.marks.length &&
-      r.marks.filter((i) => Array.isArray(i) && i[0] === 'inlineTopic').length
+      r.marks.filter(
```
pzhine, line 42: same as above

pzhine (changes requested):

PR #756, opened 2021-05-10
Michael/5 10 bugfix

msenejoa:
* https://www.notion.so/databyss/Invalid-type-number-expected-string-detail-year-textValue-still-exists-33971fec71d84cc2892bf70a2b7493b9
* https://www.notion.so/databyss/Logout-if-groupname-is-false-in-the-url-c50a916ae2814bf489cd10b729ba4394
* https://www.notion.so/databyss/Suggestion-menu-needs-to-be-fixed-You-should-be-able-to-scroll-and-see-more-than-4-Also-the-clos-c830ac38ba93400a963a8c761c2f958d
* https://www.notion.so/databyss/Turn-Off-In-Line-Topics-on-Lines-That-you-Begin-with-or-7e4ff5bb3f9a438fa146aa5eaff5fa07


jrceno:
The Source Suggestion Box should always show the reference catalogs. So we do need to limit the amount of suggestions shown just for Sources.

PR #795, opened 2021-10-31
paul/webworkers

pzhine:


pzhine:
abandoning webpack5 upgrade for now

